PR #38, opened 2015-02-16
Scrabble score

verdammelt:
@wobh I would like your review. Seems very straight forward. Maybe put in some example tests for the extra-credit?

### DIFF
--- a/config.json
+++ b/config.json
@@ -22,7 +22,8 @@
     "meetup",
     "triangle",
     "difference-of-squares",
-    "roman-numerals"
+    "roman-numerals",
+    "scrabble-score"
   ],
   "deprecated": [
 
--- a/scrabble-score/example.lisp
+++ b/scrabble-score/example.lisp
@@ -1,0 +1,20 @@
+(defpackage #:scrabble-score
+  (:use #:cl)
+  (:export #:score-letter #:score-word))
+
+(in-package #:scrabble-score)
+
+(defparameter *letter-scores*
+  '((#\A . 1) (#\B . 3) (#\C . 3) (#\D . 2) (#\E . 1) (#\F . 4)
+    (#\G . 2) (#\H . 4) (#\I . 1) (#\J . 8) (#\K . 5) (#\L . 1)
+    (#\M . 3) (#\N . 1) (#\O . 1) (#\P . 3) (#\Q . 10) (#\R . 1)
+    (#\S . 1) (#\T . 1) (#\U . 1) (#\V . 4) (#\W . 4) (#\X . 8)
+    (#\Y . 4) (#\Z . 10)))
+
+(defun score-letter (letter)
+  "Returns the score for a single letter."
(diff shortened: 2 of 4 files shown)


wobh:
Well this looks good, and I think we could merge it in as-is. I have few notes for future iterations, unless you think there are any worth doing right away:

I think "scrabble-score:score-word" makes for a bulky symbol. However, I can't come up with a better name, both package and interned-symbol parts make sense on their own. A general "scrabble" package seems the most obvious alternative. But I think a real CL scrabble game would most likely have a "scrabble-score" package, especially if they follow the "one package per file" setup. It doesn't matter to us, but I appreciate it when an exercise affects a certain amount of "realism" even though I know it's a pretense.

I think we could name these tests a bit better, a bit more like the way xruby and xelixir named their tests. They also tested the whitespace case, the empty string case, and something called "convenient score" which I don't know what the mean by that.

I kinda dislike the actual test cases everyone is using.

- "MULTIBILLIONAIRE" should be disqualified as too long. Scrabble boards are 15x15 and diagonal plays are forbidden. The rationale for it is probably that, board-dependent validations are for the extensions, although it's not mentioned in the README.
- shouldn't a test of mixed case have two alternately cased examples, eg. `(assert-equal 14 (scrabble-score:score-word "cAbBaGe") (scabble-score:score-word "CaBbAgE"))` ?
- we should find a combination of words that can uniquely score every letter to verify the user has submitted a standard scoring. Well actually since we're not validating the words we could just score "AEIOULNRST", "DG", "BCMP", etc.

I'm kind of excited about coming up with tests for the bonuses, but should we include them here or come up with a "Scrabble score 2" exercism for them (I think I'd prefer the latter).

verdammelt:
@wobh I took a moment to improve the names, added whitespace tests.

I removed score-letter. I didn't think it was really worth having in the API anymore.

I think validation of inputs (e.g. length of string) is outside the scope of scoring - so I don't have a problem with MULTIBILLIONAIRE.

I think doing bonuses would work well as a separate exercise.

kytrinyx:
Even though the problem ignores validation, I still think we should find a different word than `MULTIBILLIONAIRE` across the board. I'll add a note to the x-common repository.

verdammelt:
Here is a list of 15 character (or more) words: http://www.becomeawordgameexpert.com/wordlists15.htm. for possible future reference.

I just ran `cat /usr/share/dict/words | grep -e '^...............$'` on my macbook and have 5925 15 character words.  So here is a little script:

```
cat /usr/share/dict/words | grep -e '^...............$' > longwords
numwords=`cat longwords | wc -l`
num=$((RANDOM%numwords))
sed -n "${num},${num}p" longwords
```

And it generated: "Gastrochaenidae"

If no one has a specific favorite 15-character word that they'd really like to use - I'll pick a random word and use it.

kytrinyx:
Random is great :)

PR #40, opened 2015-03-09
Fixing Gigasecond example.

verdammelt:
The example code was wrong because it did not take into consideration
the timezone at all. Once it did that it brought to light that other
tests were incorrect.


kytrinyx:
Nice catch!

PR #45, opened 2015-04-24
Add raindrops exercise

canweriotnow:
Would appreciate thoughts from @wobh, @verdammelt et al. - `raindrops/example.lisp` is pretty derivative of my Scheme implementation, maybe could be improved to be more idiomatic CL?

### DIFF
--- a/raindrops/raindrops-test.lisp
+++ b/raindrops/raindrops-test.lisp
@@ -1,0 +1,72 @@
+(ql:quickload "lisp-unit")
+
+(defpackage #:raindrops-test
+  (:use #:common-lisp #:lisp-unit))
+
+(load "raindrops")
+
+(in-package #:raindrops-test)
+
+(define-test test-1
+    (assert-equal "1"
+                  (raindrops:convert 1)))
+
+(define-test test-3
+    (assert-equal "Pling"
+                  (raindrops:convert 3)))
+
+(define-test test-5
+    (assert-equal "Plang"
+                  (raindrops:convert 5)))
+
+(define-test test-7
+    (assert-equal "Plong"
+                  (raindrops:convert 7)))
+
+(define-test test-6
+    (assert-equal "Pling"
+                  (raindrops:convert 6)))
+
+(define-test test-9
+    (assert-equal "Pling"
+                  (raindrops:convert 9)))
+
+(define-test test-10
+    (assert-equal "Plang"
+                  (raindrops:convert 10)))
+
+(define-test test-15
+    (assert-equal "PlingPlang"
+                  (raindrops:convert 15)))
+
+(define-test test-21
+    (assert-equal "PlingPlong"
+                  (raindrops:convert 21)))
+
+(define-test test-25
+    (assert-equal "Plang"
+                  (raindrops:convert 25)))
+
+(define-test test-35
+    (assert-equal "PlangPlong"
+                  (raindrops:convert 35)))
+
+(define-test test-49
+    (assert-equal "Plong"
+                  (raindrops:convert 49)))
+
+(define-test test-52
+    (assert-equal "52"
+                  (raindrops:convert 52)))
+
+(define-test test-105
+    (assert-equal "PlingPlangPlong"
+                  (raindrops:convert 105)))
+
+(define-test test-12121
+    (assert-equal "12121"
+                  (raindrops:convert 12121)))
+
+(let ((*print-errors* t)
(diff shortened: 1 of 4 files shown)


verdammelt:
*PING* Hello? any update to this pull request?

raindrops/raindrops-test.lisp
```diff
@@ -0,0 +1,72 @@
+(ql:quickload "lisp-unit")
+
+(defpackage #:raindrops-test
+  (:use #:common-lisp #:lisp-unit))
+
+(load "raindrops")
```
wobh: `You'll need to change this to `#-xlisp-test (load "raindrops")` to run in CI. See https://github.com/exercism/xlisp/pull/54

raindrops/raindrops-test.lisp
```diff
@@ -0,0 +1,72 @@
...
+
+(define-test test-105
+    (assert-equal "PlingPlangPlong"
+                  (raindrops:convert 105)))
+
+(define-test test-12121
+    (assert-equal "12121"
+                  (raindrops:convert 12121)))
+
+(let ((*print-errors* t)
```
wobh: Add a new line above here with #+xlisp-test to run in CI. See https://github.com/exercism/xlisp/pull/54

canweriotnow:
Ah, I didn't even see comments until @wobh pinged me on gitter... I'll revisit and update this weekend, sorry!

canweriotnow:
@wobh the xlisp test reader macro has been added as required, I believe. 

@verdammelt I like the example, but I'm sure it can be improved... as the example doesn't directly affect user experience, might I suggest merging and then you (or whoever) can submit a new PR with a better example implementation?

wobh:
I'd like you to take another crack at it, as there's number of problematic things going on here. I realize it's "just" an example, but I think it's worth some extra effort (it's not much). For one thing, examples are testing the tests across multiple implementations, so we kinda need them to be solid.

Here's the problematic stuff I see, which I think you'll find interesting at least:

- There's a recursive call to `factor` inside `factor`'s `loop`. That's ... I was first surprised that didn't just blow up, but it appears to "work" because both branches of the `cond` `return` which prevents `loop` from looping.
- The recursive call isn't in the tail position so it can't be optimized in any CL implementation that might otherwise do so. Additionally, since it's in the body of a `loop` macro, even if it were in the tail of the `cond` branch, it might still not be in the tail position of macroexpanded forms generated by `loop`.
- There's no consistent way to control when TCO recursion may or may not be optimized between implementations. In CL it's best to treat recursion as a "low-level" technique and prefer "high-level" constructs until you're in a problem where it's clearly more suitable than the standard means. This isn't such a problem.
- Even with a pure `loop` factorization, we should consider, this exercise doesn't require a full factorization of the subject (or factorization at all, really), and I don't expect it ever will. But it is possible (certainly less unlikely) that we would add a largish prime or near-prime to the test cases (well, there's one already), and I think it's fair to expect that the example code should perform in constant time for that and every other case.

So here's what I recommend, use `dolist` to iterate over the required factors and associated words, use `with-output-to-string` to collect the words or numbers as they may be. Whatever you do, do the simplest thing that could possibly work.

And thank you!

canweriotnow:
Fair enough, I'll take another crack at it. No promises on timeframe,
however ;)

canweriotnow:
@verdammelt It would be very different from the examples in other languages that use prime-factorization, but what about straight modulo arithmetic instead of having the factors fn at all?

wobh:
That's pretty much what I'm suggesting. The readme text says factors, but is also alludes to fizzbuzz. And if I read the tests right, they're just require a divisibilty test.

wobh:
It's occurred to me that I'm probably going to nitpick that exactly thing a lot. I also think my ruby version of this use prime sieve. (facepalm)

kytrinyx:
I wrote the words "prime" in the README as a red herring. This is basically a re-imagining of fizz buzz, and can be solved entirely with modulo.

verdammelt:
Sorry @canweriotnow but I couldn't help myself. I wanted to try to solve this myself.  How about one of these for the example code:

https://gist.github.com/verdammelt/6f2ba3ea7cdeeea9b38e

yurrriq:
> I wrote the words "prime" in the README as a red herring.

Love it.

Also :+1: for the `loop` version. I love the `&optional` too.

verdammelt:
Superseded by #73

PR #52, opened 2015-05-27
Binary problem.

verdammelt:
I wrote several solutions - which do people think should be the 'official' one? Or maybe we leave all of them in the example file? But even in that case one should be left as the one not commented out.

### DIFF
--- a/binary/binary-test.lisp
+++ b/binary/binary-test.lisp
@@ -1,0 +1,36 @@
+(ql:quickload "lisp-unit")
+
+(defpackage #:binary-test
+  (:use #:common-lisp #:lisp-unit))
+
+(load "binary")
+
+(in-package #:binary-test)
+
+(define-test binary-1-is-decimal-1
+  (assert-equal 1 (binary:to-decimal "1")))
+
+(define-test binary-10-is-decimal-2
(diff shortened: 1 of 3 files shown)


wobh:
http://lisptips.com/post/44261316742/how-do-i-convert-an-integer-to-a-list-of-bits

wobh:
I thought I read somewhere that the example files aren't meant to be "canonical". I've been thinking of them as "meta-tests" in the sense that, if we change the testing suite, the example file will help us make sure we're still doing something reasonable. But I do see the case that making them into "good" examples would be useful for people contributing future exercises.

verdammelt:
I always thought of the example file to be "a solution" that one could look at if one got stuck. But now I don't know where the user could actually see this example file... I don't seem to get them when I do `exercism fetch`.

Thanks for the link - I never remember all the `ash` and `logand` type functions - my brain never seems to think of them (dumb brain).

kytrinyx:
The only thing the example file is used for at the moment, is as a sanity check for the test suite. It would be a shame to have an unsolvable test suite. People could technically go look at the solution in the github repo, but I doubt that many do.

Issue #26, opened 2014-12-02
Travis Integration


verdammelt:
Thought it would be good to get a real travis integration. Maybe that is possible?

- [x] Travis Badge for README
- [ ] Travis job to run tests (with sbcl for example)
  - [ ] install sbcl
  - [ ] install quicklisp
  - [ ] install lisp-unit
- [ ] solve problem of the tests loading files that do not exist (example files are named `example.lisp`)
- [ ] ???
- [ ] profit.

(*note* IMNSHO it would be cool to have the build script, once sbcl &c. are installed be written in Lisp)

wobh:
@verdammelt (cc @kytrinyx) It took me way too long to think of this, but the "dumbest thing thing that could possibly work" to get the tests to run on the example files is this: symbolic links. Our test script just has to check that the starter slug file is a symbolic link and if not, blow it away with a symbolic link to `example.lisp`. Then it runs the tests with `sbcl --load exercise-tests.lisp`.

I don't know how Travis figures out whether the lisp tests pass, but presumably it does somehow, I can look into that. Certainly there could be other drawbacks too. I'll do some local testing when I have time.

kytrinyx:
Some of the other tracks use a temporary directory and rewrite files. I don't think that matters on travis, but should anyone run the full test suite locally before committing, blowing away a stub file and having it be a symlink to the example could be problematic.

wobh:
I've come up with something better, dumber:

```common-lisp
(loop  
  for demo in (directory "./*/example.lisp")
  for test in (directory "./*/*-test.lisp")
  do (progn (load demo) (load test)))
```

verdammelt:
That won't work because the test file actually calls `load` on the code file.

wobh:
Ah, but the tests are run on the package, not the file. If we skip past the load error the tests run against the definitions in example file loaded earlier. I've been working with this for a while now and I'm now 80% sure we can hack up something that'll work.

A couple of challenges and possibilities so far:

- Clisp's file error has so-far been uncatchable. This might be a bug in clisp. However, it seems we can catch the file-error raised in ccl and sbcl.
- `*default-pathname-defaults*` is set to an empty path in ccl, but to the pathname lisp was started in in sbcl and clisp. In any case I think we can set it, in advance of loading the testfile, which should have the effect of changing the working directory for `load`. As long as we provide a stub code file, this would avoid load errors.
[...]

PR #54, opened 2015-06-02
xlisp #26: setup xlisp-test and cl-travis

wobh:
- setup bin/xlisp-test
- setup tests with xlisp-test feature
- setup .travis.yml
- extensive testing
- move robot print-object for specs
- allow failures on non-SBCL implementations for the time being

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,33 @@
-language: bash
+language: lisp
+
+env:
+    matrix:
+      # - LISP=abcl
+      # - LISP=allegro
+      - LISP=sbcl
+      - LISP=sbcl32
+      - LISP=ccl
+      - LISP=ccl32
+      - LISP=clisp
+      - LISP=clisp32
+      # - LISP=cmucl
+      - LISP=ecl
+
+matrix:
+  allow_failures:
+    # - env: LISP=abcl
+    # - env: LISP=allegro
+    - env: LISP=ccl
+    - env: LISP=ccl32
+    - env: LISP=clisp
+    - env: LISP=clisp32
+    # - env: LISP=cmucl
+    - env: LISP=ecl
+
+install:
+  - curl -L https://github.com/luismbo/cl-travis/raw/master/install.sh | sh
 
 script:
   - bin/fetch-configlet
   - bin/configlet .
+  - cl -l 'lisp-unit' -f 'bin/xlisp-test.lisp' -e '(when (xlisp-test:run-tests-all) (uiop:quit 4))'
--- a/anagram/anagram-test.lisp
+++ b/anagram/anagram-test.lisp
@@ -3,7 +3,7 @@
 (defpackage #:anagram-test
   (:use #:common-lisp #:lisp-unit))
 
-(load "anagram")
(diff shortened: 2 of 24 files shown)


verdammelt:
What is meant by "allow problem builds for the time being"?

wobh:
It should probably read "allow failures on non-SBCL implementations for the time being".

verdammelt:
Would it be possible to merge this in with *only* those implementations that work fine and open issues to fix those that don't? And have the build fail if anything is wrong?

wobh:
You mean remove the other implementations from ".travis.yml" entirely? We'd just add them back, and have to fix whatever the issues are, under the issue of "support clisp" or whatever. If you would prefer this procedurally, that's fine by me, I just want to clarify my next action and hear your rationale.

verdammelt:
I'm probably misunderstanding something.

If the build will not fail for known broken implementations (e.g. clisp) - will it fail if SBCL has a problem?

My rational is that if we are going to include an automated build with a test suite I want it to fail when there is a problem with known good environments. I would not like a build that is always failing because of known bad implementations or always passes even if there are problems.

wobh:
Ah. Check out the build page: https://travis-ci.org/exercism/xlisp

There's two SBCL environments (one's 32 bit) and the tests all pass on them.

You can see an example of failing tests on SBCL back in the build history here. I spent some time trying to build with deliberately introduced bugs to cause the tests to fail, which they do. Here's the first build where they fail as expected: 

https://travis-ci.org/wobh/xlisp/jobs/64213987

Here's the build after I removed the bugs:

https://travis-ci.org/wobh/xlisp/builds/64214537

You can see that SBCL passes while all others fail. Only thing I've done since is catch up with the recently added exercises and added the failing implementations to the "allowed failures" list. Once merged we can file the issues to fix Clisp, CCL and the others.

verdammelt:
Ah! OK, thanks for clearing up my confusion. I've got no problems with this now.

verdammelt:
(As an aside: I just scanned the CCL & ECL failures (looks the same) and for some reason 'bottles' is capitalized?! Looking at the example code I don't see anything that *might* prompt an implementation to do that... weird.)

wobh:
I've been observing a protocol where I don't merge my own pull requests. That seems slightly silly now that I type it out, but if you're ready to move on with it, either merge or let me know and I'll do it. Maybe it would be a good idea to work out some kind of signal.

verdammelt:
That seems like a fine protocol. 

Pull Request signal? http://publicsafety.tufts.edu/police/files/pubsafe_up__16.jpg

(I was hoping for something more entertaining to appear from a google image search of 'pull request signal' but *sigh* no such luck).

wobh:
Thanks! I'll file some new issues for getting the other implementations working. We should be able to check off a number of items on your checklist. Exciting times!

verdammelt:
Thanks. See my notes above about what I've debugged of CCL & ECL problems - I think if the example code did not rely upon `~(~)` it would be portable to those two implementations (and their bugs). The CLisp problem appeared easy to fix as well - at worst we remove the offending, but unused, method.

Issue #55, opened 2015-06-06
Enable testing in CI for CCL


wobh:
- [ ] fix failing spec for Beer Song https://travis-ci.org/exercism/xlisp/jobs/65655344#L966
- [ ] remove `ccl` and `ccl32` from `allow_failures` matrix in `.travis.yml`

See https://github.com/exercism/xlisp/pull/54#issuecomment-108839095

verdammelt:
Fixed by #60

PR #60, opened 2015-06-12
Fix for Issue #55 - CCL test failure in CI build.

verdammelt:
On SBCL it capitalizes only the first word. On CCL it acts like
STRING-CAPITALIZE.


verdammelt:
@wobh Thanks... I had done the commit during my lunch but didn't have time to do the pull request (or noticing that this also fixed ECL).

Issue #56, opened 2015-06-06
Enable testing in CI for Clisp


wobh:
- [x] fix error loading file "etl-test" https://travis-ci.org/exercism/xlisp/jobs/65655346#L1180
- [x] fix any other errors that emerge
  - list errors here for future reference
- [x] fix any failing specs
  - list failing specs here for future reference
- [x] remove `clisp` and `clisp32` from `allow_failures` matrix in `.travis.yml`

See https://github.com/exercism/xlisp/pull/54#issuecomment-108839095

PR #61, opened 2015-06-12
Removing offending (unnecessary) code so that Clisp will build.

verdammelt:
Fix for #56


verdammelt:
@wobh I like the idea of your protocol of not merging one's own pull requests. So if you could do the honors (if the builds do in fact pass...)

wobh:
One test timed out. Ran fine when restarted.

PR #74, opened 2015-09-28
Add atbash-cipher exercise.

verdammelt:
Tests shamelessly stolen from xclojure.

### DIFF
--- a/atbash-cipher/atbash-cipher-test.lisp
+++ b/atbash-cipher/atbash-cipher-test.lisp
@@ -1,0 +1,42 @@
+(ql:quickload "lisp-unit")
+
+(defpackage #:atbash-cipher-test
+  (:use #:common-lisp #:lisp-unit))
+
+#-xlisp-test (load "atbash-cipher")
+
+(in-package #:atbash-cipher-test)
+
+(define-test encode-no
+  (assert-equal "ml" (atbash-cipher:encode "no")))
+
+(define-test encode-yes
+  (assert-equal "bvh" (atbash-cipher:encode "yes")))
+
+(define-test encode-OMG
+  (assert-equal "lnt" (atbash-cipher:encode "OMG")))
+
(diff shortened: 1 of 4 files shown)


kytrinyx:
... as one should :)

wobh:
Ubiquity of Unicode notwithstanding, but I would feel uncomfortable using `char-code` this way, for things like this. I know it makes me weird. My idiosyncrasies notwithstanding, I'm merging. Nice work!

verdammelt:
1. I believe that this cipher might date from Roman times - my algorithm handles more characters than they had  :)

2. The Common Lisp spec specifies that A>Z and a>z and 0>9 everything else is up to the implementation. So my code is implementation independent.

Those are weak arguments for not doing something with Unicode - but that's my story and I'm sticking with it!

(thanks for merging).

wobh:
Because pedantry is my favorite sport (although I hate that I like it so much):

1. I believe it's at least a few hundred years older than the Roman empire.

2. The Common Lisp spec does not specify the order of characters that an implementation may use for`char-code`, only that it have one. The successor of `#\m` may be `#\M`, or `#\N`, or `#\n`. That's what you're counting on unicode (and the legacy of ascii) for. It's a super safe bet, but not an implementation independent one. For that I recommend `digit-char-p` with the optional radix argument of 36.

Those are weak arguments for fixing it. To do something with Unicode we'd need a reference for all Unicode alphabets that have some kind of expected total order that the Unicode code points coincidentally correspond to. I don't think even Elixir has that.

(of course!)

verdammelt:
Oooh I love me some good pedantry (especially when I learn things).

I'l look at the spec again about `char-code` and ordering. I must have read it wrong.

The alternative I had thought of was to have a sequence of characters in alphabetical order and another reversed. Look up your character in one to find its index then use that in the second to find the ciphered letter.

Or, I just thought of using an alist for the same.

PR #75, opened 2015-09-28
Add new accumulate exercise.

verdammelt:
Very simple exercise - perhaps we should make it one of the first.

### DIFF
--- a/accumulate/accumulate-test.lisp
+++ b/accumulate/accumulate-test.lisp
@@ -1,0 +1,36 @@
+(ql:quickload "lisp-unit")
+
+(defpackage #:accumulate-test
+  (:use #:common-lisp #:lisp-unit))
+
+#-xlisp-test (load "accumulate")
+
+(in-package #:accumulate-test)
+
+(defun square (x) (* x x))
+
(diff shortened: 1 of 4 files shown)


verdammelt:
Maybe we shouldn't include this one at all. It is really simple due to `mapcar`.

Or maybe we make it more interesting by having it take different types of sequences (returning a sequence of the same type)? what would allow the use of `typecase` or generic functions.

yurrriq:
Or can we add "don't use `mapcar`" or even `map*` to the README somehow? There are some fun ways to implement this "manually." :+1: for making it the first exercise too.

verdammelt:
It seems that @canweriotnow added a note to the readme (https://github.com/exercism/x-common/blob/master/accumulate.md) stating that `mapcar` was just fine.

I'm leaning toward making this a "foregone" exercise. I don't think we should include it.

yurrriq:
Cool. :+1: for exclusion then.

wobh:
Agree we should exclude it.

verdammelt:
I'll change this pull request to be an exclusion and then merge it in.

Issue #23, opened 2014-12-01
Implement Lisp Exercises


verdammelt:
This is a placeholder issue for keeping track of porting assignments to Lisp, or noting why they may not be applicable. (This list shamelessly stolen from https://github.com/exercism/xclojure/issues/1)

- [X] bob
- [X] word-count
- [X] anagram
- [X] beer-song
- [X] nucleotide-count
- [X] rna-transcription
- [X] point-mutations
- [X] phone-number
- [X] grade-school
- [X] robot-name
- [X] leap
- [X] etl
- [X] meetup
- [X] space-age
- [X] grains
- [X] gigasecond
- [X] triangle
- [X] scrabble-score
- [X] roman-numerals
- [X] binary
- [x] prime-factors
- [X] raindrops 
- [X] allergies
- [X] strain
- [X] atbash-cipher
- ~~accumulate~~
- ~~bank-account~~
- [X] crypto-square
- [ ] trinary @verdammelt
- [x] sieve
- [ ] simple-cipher
- [ ] octal
- [ ] luhn
- [ ] pig-latin
- [ ] pythagorean-triplet @wobh
- [ ] series
- [X] difference-of-squares
- [ ] secret-handshake
- [ ] linked-list
- [ ] wordy
- [ ] hexadecimal
- [ ] largest-series-product
- [ ] kindergarden-garden
- [ ] binary-search-tree
- [ ] matrix
- [ ] robot-simulator
- [ ] nth-prime @wobh
- [ ] palindrome-products
- [ ] pascals-triangle @wobh
- [ ] say
- [ ] sum-of-multiples
- [ ] queen-attack
- [ ] saddle-points
- [ ] ocr-numbers

wobh:
How do you want to divvy these up? There's a number of simple ones, we should probably try to get out sooner rather than later: roman numerals, difference of squares, nth-prime, pascals triangle, raindrops, and probably others. They can be cycled earlier into the list of existing exercises.

verdammelt:
Maybe we can just note down when one of us 'takes' one to do it. Then handle the submission via pull request?

wobh:
Got "difference of squares" almost ready. Roman numerals next.

canweriotnow:
PR #45 should check off 'raindrops'

wobh:
[...]

PR #79, opened 2015-10-01
Xlisp #23 prime factors

wobh:
Merge after xlisp-77 PR.


wobh:
One note: I added a bunch of tests based on some composites that make primality testing tricky. Most of the tests in other tracks seemed to be concerned with factoring early prime powers, and I thought we should step up the game a bit. I included those other tests although, I think most of them are obviated by the new ones.

PR #80, opened 2015-10-18
Foregoing bank-account exercise.

verdammelt:
The exercise specifically talks about threads which cannot be done in
an implementation independant fashion in Common Lisp. Removing that
from the exercise does not leave much of interest that has not been
seen already.

### DIFF
--- a/config.json
+++ b/config.json
@@ -39,6 +39,7 @@
     "docs"
   ],
   "foregone": [
-    "accumulate"
+      "accumulate",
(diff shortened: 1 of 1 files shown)


wobh:
Agreed. I'll merge a bit later this evening to give more time if someone objects.

yurrriq:
:+1:

kytrinyx:
:+1:

PR #85, opened 2015-11-20
Added test for the uniqueness of robot names

TatriX:
Test will check that *most* of the names are unique. It actually *could* check that all of the names are unique but due to large amount of names most of the user implementations will be slow. Anyway most of the time test will detect non-unique names.

### DIFF
--- a/robot-name/example.lisp
+++ b/robot-name/example.lisp
@@ -4,20 +4,38 @@
 
 (in-package #:robot)
 
-(defun random-alpha-char ()
-  (code-char (+ (char-code #\A) (random 26))))
-(defun random-digit-char ()
-  (code-char (+ (char-code #\0) (random 10))))
+(defparameter *robot-names* (make-hash-table :test #'equal))
 
-(defun random-robot-name ()
-  (concatenate 'string (list (random-alpha-char)
-                             (random-alpha-char)
-                             (random-digit-char)
-                             (random-digit-char)
-                             (random-digit-char))))
-
-(defun build-robot () (make-instance 'robot))
 (defclass robot ()
-  ((name :reader robot-name :initform (random-robot-name))))
+    ((name :initarg :name :accessor robot-name)))
+
+(defun build-robot ()
+  (make-instance 'robot :name (unique-random-name)))
+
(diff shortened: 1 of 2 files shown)


verdammelt:
Congratulations! It looks like your test fails. :)

Now, what can we do to make it pass? The build runs all tests against the code in `example.lisp` in each exercise. And it does this on ~6 implementations. (SBCL, CMUCL, ABCL etc).

TatriX:
If I understand correctly it failed because I forgot to make ```example.lisp``` to pass new test itself.
And here is the problem: my implementation of example, which randomly picks names until it's unique, is *very* slow when there are few names left. I've tried to make it run in constant time in several ways but I didn't like them all.

I've found one implementation which used custom random number generator but I thinks it's a bad example for newbies ;)

wobh:
I don't think you have to exhaust the robot factory to adequately test uniqueness of name generation. Instead, start generating robots, but randomly (1 in 5?), turn off a robot instead of (or even in addition to, since that's safer) making one. When the population of activated robots reaches some significant sample size (100?), check that they have unique names.

wobh:
It occurs to me that for statistical tests we'll want a way of marking them so they either don't run in all test runs, or don't cause a build failure if they happen to fail. This would smooth the way for adding randomness and uniqueness tests or other statistical tests for large domains.

TatriX:
Well, now test failed in clisp due to max execution time limit of 3 seconds I made for uniqueness test. Of course I can simply increase that time or decrease number of iterations.

But first I want to know what do you think about my approach in general.

verdammelt:
What if you just generated robot names N times (where N is big, but no larger than `(* 26 26 1000)`, use `remove-duplicates` and check if you still have N names. Would that help the performance of the test?

TatriX:
The problem with current test is not the test itself but example.lisp which randomly trying to generate unique names. So when the more names generated the slower algorithms works. I think it's fine for this particular exercise but I will be happy to use another approach.

verdammelt:
Build never fixed in 2 months.

I'm going to close this.

PR #88, opened 2015-11-28
Preparing Cryptosquare exercise.

verdammelt:


wobh:
This looks fine. I'm prepared to merge it, but I wonder if there's a simpler example implementation we should consider. I rewrote the core functions using `dotimes` and `rotatef` which I'm providing here for informational purposes--not recommending as I don't think they're any better.

```common-lisp
(defun ntranspose (matrix)
  "Destructively transpose elements of a square array."
  (assert (= 2 (array-rank matrix)))
  (assert (= 1 (length (remove-duplicates (array-dimensions matrix)))))
  (let ((dim (first (array-dimensions matrix))))
    (dotimes (x (- dim 1) matrix)
      (dotimes (m (- dim x 1))
        (let ((y (+ m (1+ x))))
         (rotatef (aref matrix x y)
                  (aref matrix y x)))))))

(defun text-to-matrix (text)
  "Convert string to square array."
  (let* ((n (ceiling (sqrt (length text))))
         (m (make-array (list n n) :initial-element #\Space)))
    (dotimes (i (length text) m)
      (setf (row-major-aref m i) (char text i)))))

(defun matrix-to-text (matrix)
  "Convert square array of characters to string."
  (with-output-to-string (str)
    (dotimes (i (array-total-size matrix))
      (let ((ch (row-major-aref matrix i))
            (dim (array-dimension matrix 0)))
        (when (and (plusp i) (zerop (mod i dim)))
          (princ #\Space str))
        (when (char/= ch #\Space)
          (princ ch str))))))
```

verdammelt:
Using `dotimes` would be fine too. I had trouble making this example anything but big and ugly looking (but still working well). I'm hoping to come up with a nicer or at least more interesting solution when I solve it as a submitter.

So: my solution with `do` or your solution with `dotimes` are both fine by me.

What is your reasoning for the destructive transpose function instead of a non-destructive one? "performance"?

wobh:
"Reasoning" is too strong a word. It's a gesture toward the ideal of simplicity. In retrospect, the transpose operation is reversible, the matrix is an intermediary structure, and both those make it seem like a case for destructive operations would be fine. I think we could probably do without the intermediary array and work on the normalized string. I think a fairly direct implementation from input string to output string should be possible. Let's see if someone implements something we like and ask them to PR it.

verdammelt:
I like that idea a lot.

PR #92, opened 2015-12-07
Get off the legacy platform on Travis CI

kytrinyx:
We don't need sudo, so we should be able to use the containerized build infrastructure.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,7 @@
 language: lisp
 
+sudo: false
+
 env:
     matrix:
       - LISP=sbcl


kytrinyx:
I guess not :( I'm lost as to how to get this working.

arguello:
The install script uses `sudo`:
https://github.com/luismbo/cl-travis/raw/master/install.sh

That might be causing the errors.

I was able to get the checks to pass by modifying the install script: https://github.com/exercism/xlisp/pull/93

We could also try removing sudo from the commands in the install script, and test the containerized build that way. I'm not too familiar with the containerized build infrastructure, though.

kytrinyx:
Ah, I completely missed that. Good catch. I'm going to close this--my lack of familiarity with the build is going to make more work for y'all than it's worth :)

PR #91, opened 2015-12-07
Normalize YAML in Travis config

kytrinyx:

Issue #58, opened 2015-06-06
Update contributing documentation with info about using CIM


wobh:
Basically,
1. Install CIM
2. Install lisps
3. Install Quicklisp
4. --Run all the tests locally--

See https://github.com/KeenS/CIM

verdammelt:
I had problems with CIM in that I already had QuickLisp installed and CIM didn't seem to see it. As far as I can tell CIM expects a quicklisp directory inside of $CIM_HOME. This seems awfully presumptuous of it :).

I made it work by symlinking my ~/.quicklisp to ~/.cim/quicklisp.

So now that I've solved that mystery I *could* write up the instructions - but perhaps instructions on how to install CIM/Lisps/QuickLisp in the 'normal' way might work better?

For the record: I installed SBCL via homebrew, QuickLisp, then CIM, then more implementations via homebrew. I am guessing this is not the 'correct' way.

Also note: the item about how to run items locally is handled already by my changes to the README.

yurrriq:
Mildly related questions:
- Is there a CIM equivalent for scheme? I often find I need to switch between chicken, gambit, guile and racket.
- Is there a good way to get gambit and Ghostscript to coexist peacefully?

verdammelt:
Been pretty quiet on this issue lately. If no one else wants to write this up I'll pick this up and try to figure out the 'right' way of installing CIM.

verdammelt:
I finally got around to trying to use CIM directly and boy am I now annoyed. Maybe we shouldn't suggest CIM? It may not be worth it... have other people had good luck with it? 

Just now it would not install clisp-2.49 (had to install clisp-head). There is no update option for the lisps it installs. it does not consistently install quicklisp in all lisps. I purged all installs and tried to re-install sbcl via brew (which succeeded) but then coulnd't get CIM to work properly with sbcl-system. grrr.

yurrriq:
I've had similar experiences, @verdammelt :/

wobh:
Okay, lets look at our options.

verdammelt:
Funny thing is CIM seems to work nicely when I had it set up. I had it set up with all HomeBrew installed implementations and with QuickLisp installed by hand and symlinked into the CIM directory. That let me use it as we do for testing. But trying to use it as a Lisp Implementation manager system ala RVM/RBENV, etc. etc. I don't think that's going to work.

[...]

PR #93, opened 2015-12-07
test install script

arguello:
Testing Travis CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ matrix:
       - LISP=abcl
 
 install:
-  - curl -L https://github.com/luismbo/cl-travis/raw/master/install.sh | sh
+  - curl -L https://github.com/arguello/cl-travis/raw/master/install.sh | sh
 
 script:
   - bin/fetch-configlet


arguello:
The tests passed - all I did was modify the install script:
https://github.com/luismbo/cl-travis/compare/master...arguello:master

I'm not sure if this is the best way, and a little concerned that this broke in the 1st place - but it works.

wobh:
Yeah, this seems like a known thing: https://github.com/luismbo/cl-travis/issues/9. I'd prefer to try to get it fixed upstream than fork. I'm intrigued by the last comment there about alternatives. Maybe we should look into that...

arguello:
+1 for an upstream fix.

This test had nothing to do with sudo, though - I just added the `-y` flag to `apt-get` - gets around the continue prompts. Applying that upstream might cause it to break for others?

Looks like someone had an idea for native lisp support, but the project seems to have died down:
http://blog.cddr.org/2015/07/28/native-travisci-support-for-common-lisp/
https://github.com/lisp-travis

Couldn't find any other alternatives to cl-travis.

I'm willing to help with whatever is decided.

verdammelt:
In the issue @wobh referred to above the author of cl-travis says "folks are working on better alternatives to cl-travis.". Any ideas what they mean by that? If we don't know I, or someone else, ough to ask...

wobh:
One thing is that cl-travis installs lisps itself as system lisps, only using CIM to set the default lisp. (for example: [107: install_abcl](https://github.com/luismbo/cl-travis/blob/master/install.sh#L107), and [116: cim use abcl-system --default](https://github.com/luismbo/cl-travis/blob/master/install.sh#L116)). I've only just started looking into available features on TravisCI

- https://docs.travis-ci.com/user/languages/community-supported-languages/
- https://docs.travis-ci.com/user/apt/

and correlating those with cl-travis and CIM (see https://github.com/KeenS/CIM/blob/master/scripts/cim_lib_install), and I'm starting to wonder if we even need cl-travis? What required that cl-travis install lisps as system lisps?

arguello:
Would this be a good alternative? [Using Travis CI with Roswell] (https://gist.github.com/fukamachi/52dce85caf477043ffeb)

verdammelt:
Roswell looks like it could replace CIM. But something about the fact the project is also a CL web framework - not just a implmentaiton switcher, seems odd to me.

What if we fork CIM and use that? Then when/if CIM integrated the fix we can switch back to CIM.

arguello:
> What if we fork cl-travis and use that? Then when/if cl-travis integrated the fix we can switch back to cl-travis.

Sounds like a good idea.

We can move the `sudo apt-get` commands out of the install script:
https://docs.travis-ci.com/user/customizing-the-build/#Installing-Packages-Using-apt

Is it necessary to install both the 32-bit and 64-bit versions of each lisp?

wobh:
I'll try to setup xlisp-travis this weekend.

arguello:
:tada:

verdammelt:
@wobh good luck! If I can help let me know.

wobh:
So, it turns out the problem was that I committed a wonkily formatted travis.yml that only caused problems when the YAML directive separator [---](http://www.yaml.org/spec/1.2/spec.html#id2760395) was added to the top of the file. Once the file was reindented, the builds worked without any need to make changes to cl-travis.

arguello:
So, those jobs ran on the legacy infrastructure, and all passed. Half of the jobs that ran on the "new platform for Legacy Precise builds" failed, and I think every job that ran on the containerized build infrastructure failed.

Do the yaml updates guarantee that the jobs will always run on the legacy infrastructure?

wobh:
I think they will until TravisCI requires us to add `sudo: true` to them. https://docs.travis-ci.com/user/migrating-from-legacy

I'm still working on xlisp-travis, I just noticed that the we could resolve/merge #91 independently.

verdammelt:
@wobh good catch on the YAML formatting thing.

arguello:
Awesome - good catch for sure.

wobh:
So my plan was to to try to come up with an install script that would use cim to install lisps. This would let us get away from cl-travis's use of `sudo` and thus get us on with containers. However I've had problems working with cim. Locally, I get a build error with `cim install clisp`, and `cim install ccl` doesn't seem to work at all. Furthermore, I'm not sure cim supports installing x86 versions. We can get around this. We could drop support for 32-bit platforms. This has taken more time than I hoped it would, and I've had less time than I hoped to take with it.

verdammelt:
@wobh As you may have seen in #58 I had the same problems with CIM. I don't think it is up to the task of installing Lisps.

What is the state of this pull request?

wobh:
I'll close it. I think it was a good idea, but we're trying to fix something that's not really broken, just somewhat undesirable. At this remove, it seems like the thing to do is support the effort for native CL on TravisCI.

PR #100, opened 2016-02-14
Fixing failing test.

### LINKED ISSUE #99: Build failing after #98
#98 is a much needed update to the README, so this is kind of a WTF.

~~Failing test~~ Error in Robot, https://travis-ci.org/exercism/xlisp/jobs/109074111#L907

```
INFO: Running tests for #<PACKAGE "ROBOT">
; in: LAMBDA ()
;     (LISP-UNIT:ASSERT-TRUE
;      (AND (= (LENGTH ROBOT-NAME-TEST::NAME) 5)
;           (EVERY #'ROBOT-NAME-TEST::IS-UPPER-ALPHA-P
;                  (SUBSEQ ROBOT-NAME-TEST::NAME 0 2))
;           (EVERY #'ROBOT-NAME-TEST::IS-DIGIT-P
;                  (SUBSEQ ROBOT-NAME-TEST::NAME 2 5))))
; --> LISP-UNIT::EXPAND-T-OR-F LET 
; ==>
;   #'AND
; 
; caught ERROR:
;   The macro name AND was found as the argument to FUNCTION.
; 
; compilation unit finished
;   caught 1 ERROR condition
 | Execution error:
 | Execution of a form compiled with errors.
Form:
  #'AND
Compile-time error:
  The macro name AND was found as the argument to FUNCTION.
 |
NAME-MATCHES-EXPECTED-PATTERN: 0 assertions passed, 0 failed, and an execution error.
```

EDIT: ~~And this is weird~~

https://travis-ci.org/exercism/xlisp/jobs/109074111#L1489

EDIT: removed SBCL warnings about undefined functions in the atbash-cipher example. This is normal since SBCL warns about this if the reference comes before the definition in the file. We may want to specify an example style guideline about this but it's not a problem. From https://travis-ci.org/exercism/xlisp/jobs/109074111#L1554 on down the atbash tests run fine.

verdammelt:
Due to recent changes to lisp-unit this test was failing. lisp-unit
now expects that the argument to ASSERT-TRUE will be a function call
expression.

Fixes #99

### DIFF
--- a/robot-name/robot-name-test.lisp
+++ b/robot-name/robot-name-test.lisp
@@ -14,10 +14,11 @@
 (defparameter *clutz* (robot:build-robot))
 
 (define-test name-matches-expected-pattern
-  (let ((name (robot:robot-name *robbie*)))
-    (assert-true (and (= (length name) 5)
-		      (every #'is-upper-alpha-p (subseq name 0 2))
-		      (every #'is-digit-p (subseq name 2 5))))))
+  (flet ((name-has-correct-pattern (name)
+           (and (= (length name) 5)
+                (every #'is-upper-alpha-p (subseq name 0 2))
+                (every #'is-digit-p (subseq name 2 5)))))
+    (assert-true (name-has-correct-pattern (robot:robot-name *robbie*)))))
 
 (define-test name-is-persistent
   (assert-equal (robot:robot-name *robbie*) (robot:robot-name *robbie*)))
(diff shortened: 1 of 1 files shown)


robot-name/robot-name-test.lisp
```diff
@@ -14,10 +14,11 @@
 (defparameter *clutz* (robot:build-robot))
 
 (define-test name-matches-expected-pattern
-  (let ((name (robot:robot-name *robbie*)))
-    (assert-true (and (= (length name) 5)
-		      (every #'is-upper-alpha-p (subseq name 0 2))
-		      (every #'is-digit-p (subseq name 2 5))))))
+  (flet ((name-has-correct-pattern (name)
```
wobh: I might have misread the docs, but have you tried `(assert-true (= (length name 5) (every ...) ...)` ?

verdammelt:
I don't think what you are suggesting will work. It looks like you suggest removing the `AND` - without it then there would be 3 forms passed to `ASSERT-TRUE`. It looks like that macro takes a test argument and a `&rest` 'extras' argument. The extras are used if the `test` fails. So that is not we want - we want to check that all 3 forms evaluate to true.

Issue #104, opened 2016-02-21
"point-mutations" is deprecated in favor of hamming


kytrinyx:
This happened a while back, and it was for really weird legacy reasons.

I've since fixed the underlying issues that caused the problem, but for consistency
it would be nice to rename point-mutation to hamming, so that all the tracks are using
the same exercise name.

Once the problem has been renamed, I can run a script on the website to point people's
existing point-mutations solutions to the new hamming exercise so that they'll be able
to review solutions to hamming, and people who solve the new hamming exercise can see
all the old ones.

verdammelt:
So at first glance we'll need to:

* change the production code file name from `dna.lisp` to `hamming.lisp`
* likewise change the production code package
* change the test code package from `point-mutation-tests` to `hamming-test`
* change the name of the directory
* change config.json

@kytrinyx Has any of the description of the problem changed? All the tests we have are faux DNA sequences. Also the main function has a documentation string mentioning DNA, as well as arguments (unimaginatively) named: `dna1`, `dna2`.

kytrinyx:
No the description of the problem hasn't changed, nor has the inputs/outputs (faux DNA is fine, and dna1, dna2 sounds perfectly reasonable).

wobh:
Uh oh. Looks like I needed to work on this a bit sooner? http://exercism.io/submissions/a04719ddd9f34322805ca1dd91ea143f

What if I do it now?

verdammelt:
Hmm... didn't realize there was a deadline. I thought those submissions looked familiar :/

kytrinyx:
No, that is totally my bad. I completely and utterly screwed up. I'll see if I can sort this out today (give me an hour or so).

kytrinyx:
OK, I think this should be tidied up.

wobh:
Changing the production package name means updating the tests to use `hamming:hamming-distance`. I would prefer `hamming:distance`. However  I find myself compelled to generalize a bit, as though the demonstration problem is one about DNA (a test metaphor, if you will) and `hamming` is just a package it uses (it could also have a `hamming:weight`. PR coming soon.

kytrinyx:
`hamming:distance` is great. Any way we can make it so?

wobh:
Made so. The PR is how I would prefer it, after I did some additional tidying up.

[...]

PR #105, opened 2016-03-08
xlisp #104: renamed `point-mutations` to `hamming`

wobh:
- changed the test code package from point-mutation-tests to hamming-test
- changed the production code file name from dna.lisp to hamming.lisp
- changed the production code package
- changed the name of the directory
- changed entry in config.json


verdammelt:
Hmm failing build is a bit unexpected. And not failing the same implementations each time. The couple I've looked at shows a problem at the beginning installing the implementation.

wobh:
Blocking until #106 can be dealt with.

PR #107, opened 2016-03-24
Add Trinary exercise.

verdammelt:


verdammelt:
So I don't think that this exercise is actually worth it. It is *just* like `binary`. I grabbed the code from `binary:to-decimal` and changed `2` to `3`. (there was also an additional test that caused me to change the code a _smidge_. Then I just implemented it with `PARSE-INTEGER`.

I think maybe we drop this exercise and add that additional test to `binary`. Or just drop this exercise and leave it at that.

(cc: @exercism/lisp - what have the other lispy tracks done with `binary` and`trinary`? Is it as simple as it is in Common Lisp?)

yurrriq:
I agree they're ire similar. Octal is too. I found it kind of fun to reimplement my solutions more generically in other languages, but I'm not sure that warrants have both (or all three). :+1: for dropping trinary and adding the test.

verdammelt:
The build failures are the expected failures so things are OK.

(ugh. feel dirty just saying that :/)

verdammelt:
*IFF* we keep trinary - should we put it right after binary?

yurrriq:
I think not. It was fun for me to see it crop up down the road and then remember that I'd basically solved it before. If they were right next to each other, that'd spoil it.

wobh:
I think the idea in both exercises is to get experience parsing input streams and strings. If we used unfamiliar numeric formats, we'd have to do these exercises differently. What if we used a balanced ternary numeration for this one? For binary, what if we used "ox" (or "nu" or "mw" or "NZ" or ...) numerals?

verdammelt:
@wobh - not sure what you mean. Do you mean (for example): use 'o' and 'x' instead of '0', and '1' as the binary digits?

wobh:
Yep. Then it becomes a parsing exercise, which is what I think they're after with these ones.

kytrinyx:
I really like the idea of using something other than digits for these.

Update: Wow, yes. I really, really like it.

verdammelt:
With reference to #111 how about: 

* make `binary` have those additional tests which makes it a bit more than just `parse-integer`
* we change trinary to not use numerals as the base 3 digits which also makes more than just `parse-integer` - but in a different way.

?

wobh:
Reviewing the conversation and PR, and I think we should merge this and see how it goes. I don't think it's bad to have a few easy ones in the track, here and there. Think of it as interval training. I'd like consider exercises with alternative numerations separately and introduce them when we're ready to compare and get feedback on them.

yurrriq:
:+1:

Issue #106, opened 2016-03-08
Builds failing for 32-bit platforms


wobh:
#93 Part 2: Integration Boogaloo.

I think the culprit is in the `install_i386_arch` function. Making a branch to test this out in.

wobh:
The error message is:

    W: Failed to fetch http://dl.google.com/linux/chrome/deb/dists/stable/Release  Unable to find expected entry 'main/binary-i386/Packages' in Release file (Wrong sources.list entry or malformed file)

Which occurs in `install.sh:33` in `install_i386_arch` where `sudo apt-get update -qq` Sure enough, if you look in http://dl.google.com/linux/chrome/deb/dists/stable/Release directly, there are only `main/binary-amd64/Packages` entries. Looking into what `dpkg` options we may need now.

Here's the contents Release file
[Release.txt](https://github.com/exercism/xlisp/files/164250/Release.txt)

I note it's from Google, and not Ubuntu directly. I don't know if that's a Travis, Google, or Ubuntu thing. It's been a while since I worked in Debian, but there's a sources file which may specify this location for the packages. I'll look into if there's a Travis option for specifying this.

wobh:
Not clear to me why the allowed failures are still failing: 

- https://travis-ci.org/wobh/xlisp/jobs/114691646
- https://github.com/wobh/xlisp/tree/xlisp-106-builds-failing-for-32-bit-platforms
- https://raw.githubusercontent.com/wobh/xlisp/xlisp-106-builds-failing-for-32-bit-platforms/.travis.yml

Something about the YAML syntax?

verdammelt:
So... what's the verdict?

wobh:
No verdict, yet. I'll circle back on this ASAP. I've been crazy busy.

verdammelt:
> ...crazy busy

Oh I know that feeling - fair enough. Come back to it when you can.

wobh:
I took a look through some of the `.travis.yml` files in other tracks; xhaskell and xgo were the most informative. xgo is allowing a build to fail, however, Go has native travis support which I think lets it do yaml key values instead of a list item with environment variable literally in there. xhaskell's setup is more like ours with literal environment variable setting, but they don't allow any failing builds.

[...]

PR #113, opened 2016-04-08
xlisp #106: allow 32-bit builds to fail

wobh:
- copy and use local install script
- make install script verbose
- add arch check, abcl
- use proper key in allow_failures

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,27 +3,34 @@ language: lisp
 
 env:
   matrix:
-  - LISP=sbcl
-  - LISP=sbcl32
-  - LISP=ccl
-  - LISP=ccl32
-  - LISP=ecl
-  - LISP=clisp
-  - LISP=clisp32
-  # - LISP=allegro
-  # - LISP=cmucl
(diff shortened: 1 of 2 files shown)


wobh:
This brings in the installation script from `https://github.com/luismbo/cl-travis/raw/master/install.sh` which I'm fine with for now, even though, I kinda think it means we have to fix it. We can always set the travis config back if a real improvement happens upstream.

wobh:
PR for #104 to follow this one.

README.md
```diff
@@ -2,7 +2,7 @@
 
 Exercism exercises in Lisp
 
-[![Build Status](https://travis-ci.org/exercism/xlisp.png)](https://travis-ci.org/exercism/xlisp)
+[![Build Status](https://travis-ci.org/wobh/xlisp.png)](https://travis-ci.org/wobh/xlisp)
```
verdammelt: @wobh could you please change this back? Other than that - merge this PR.

wobh: Doh! Thank you!

PR #114, opened 2016-04-08
xlisp #104: renamed `point-mutations` to `hamming`

wobh:
Merge #113 first.


verdammelt:
@wobh now that the build thing is sorted - will you clean this one up and we'll merge it?

Issue #126, opened 2016-12-06
Relative load


spacebat:
Because there is no asd file, I find myself going to the test file to load it first. But in my emacs/slime environment, where `*default-pathname-defaults*` is likely my home directory or elsewhere, this form fails:

`#-xlisp-test (load "...")`

I've been changing that load form to: 

`(load (merge-pathnames "..." (or *load-pathname* *default-pathname-defaults*)))`

If this seems reasonable I'm happy to make a PR.

Also, nothing seems to provide the `:xlisp-test` feature - am I missing something?

verdammelt:
* `xlisp-test` feature is provided during our builds. In our build we 'manually' load `example.lisp` and then the test file. 
* I think changing the load line to something like `(load (merge-pathnames "file" *load-pathname*))` would be good.

spacebat:
PR submitted

wobh:
The load error is "by design" for better or worse. It reminds me to use `,cd` in slime to change the default directory, then, of course, loading works. I probably didn't even think to document that. Sorry.

Setting up to load the merged path is a sensible improvement. I would only suggest that, we also move the load expression to line two or three---beneath `ql:quickload`. That at least makes it clearer that this an external requirement and not really part of the test package.

For example:

```lisp
(ql:quickload "lisp-unit")
#-xlisp-test (load (merge-pathnames "raindrops"
                                    (or *load-pathname*
                                        *default-pathname-defaults*)))

(defpackage #:raindrops-test
  (:use #:common-lisp #:lisp-unit))

(in-package #:raindrops-test)

...
```

verdammelt:
@wobh Organizing the loading and the package shenanigans is a good idea.

PR #127, opened 2016-12-07
Relative load in test files for code under test

spacebat:


verdammelt:
Would a simpler `(merge "..." *load-pathname*)` be sufficient?

spacebat:
Unfortunately if merge-pathnames is passed NIL as its second argument you get an error, and `*load-pathname*` is nil unless loading is in progress. To make your suggestion friendly to interactive evaluation, we'd have to wrap it in an `(eval-when (:load-toplevel :compile-toplevel) ...)` form, which I try to avoid where possible.

verdammelt:
Sorry I didn't just test it out myself before asking. 

I don't like, what I feel is, the ugliness of the expression. I'm thinking of the experience of the person new to Common Lisp coming to the track - will this expression leave the person with the feeling "Wow - Common Lisp is really complicated" or will they just ignore it as some ritualistic boilerplate?  (But in the latter case - will they start copying it everywhere?

I guess the ugliness was already started with the feature check...

As for utility function use in other tracks - I'll check and repot back.

verdammelt:
@spacebat can you explain your workflow - because in my testing I'm not sure if I'm seeing the issue. Perhaps because the way I use my environment (SBCL & SLIME) is different.

From the issue #126 you mention that you set `*default-pathname-defaults*` to some location. In my checking no matter what my `*default-pathname-defaults*` is set to if I set my working directory in SLIME (via `cd`) it sets `*default-pathname-defaults*`. I don't normally set the defaults so either I'm starting SLIME in the correct directory or changing to the correct directory.

I want to keep the code that exercism submitters see as simple and clear as possible; mostly they are *not* experienced Common Lisp users. You appear to be an experienced Lisp user (thanks for submitting to the track!) and I want to be careful that we don't make changes to help the expert at the detriment of the beginner.

spacebat:
@verdammelt I just run M-x slime-load-file and use ido to browse to the test file (which might be the default if I have run the command with the test file in the current buffer), which then fails because I haven't run slime-cd first. My goal is to make the test file just work however it is loaded or evaluated, but I understand your concern - its more code and doesn't look approachable for new users. If we can't easily create an xlisp library for this sort of utility, perhaps the instructions should just mention the need to need to run cd in slime or start lisp in the exercise directory - I haven't found that mentioned anywhere but could have missed it.

Thanks for thinking about it though :)

verdammelt:
I think if we created some xlisp library with a helper function for loading the 'production' code - then we'd still have the problem of how to load that library :)

Unless we push it to quicklisp - but that seems over-the-top.

Let me review the README - maybe we can put some documentation about SLIME usage.

spacebat:
Lets update the docs then. I'll close this PR.

PR #128, opened 2016-12-09
Protect quoted list from mutation

spacebat:
Sort is a potentially destructive operation; this made the tests fail when run from the slime repl, yet pass when run as a side-effect of loading the file. Some head-scratching ensued :) I'm copying both sequences for symmetry and simplicity of interface.


verdammelt:
Good catch! thanks.

PR #129, opened 2016-12-14
Add missing test to anagram

spacebat:
"foo" is not an anagram of "Foo"

PR #136, opened 2017-02-26
Add word-is-not-own-anagram-case-insensitively

verdammelt:
My implementation of anagrams-for suffered from this bug.


verdammelt:
This is a rebasing of #129

Issue #143, opened 2017-03-04
Set topics on exercises


verdammelt:
ref discussion in #121 

For reference List of Topics: https://github.com/exercism/x-common/blob/master/TOPICS.txt

PR #148, opened 2017-05-09
gh #143: set topics on exercises

wobh:
I tried to add at least one topic from each of the three categories.

### DIFF
--- a/config.json
+++ b/config.json
@@ -7,147 +7,232 @@
     {
       "slug": "hamming",
       "difficulty": 1,
-      "topics": []
+      "topics": ["Control-flow (conditionals)",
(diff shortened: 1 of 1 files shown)


verdammelt:
This looks great. But just before I merge it a question: "Floating point numbers" on things like triangle? I'm not remembering any exercises so far that do much with non-integers. I'll review the exercises a bit today as well.

wobh:
I did that because there's no generic "numbers" label.

verdammelt:
Ah. Oh those silly computer languages without fuller taxonomies of Numbers.

I'm going to merge this. Thanks.

Issue #149, opened 2017-05-10
TravisCI tests not running.


wobh:
See https://travis-ci.org/exercism/xlisp/jobs/230592144

When we updated the format of the `config.json` we didn't update the way the list of exercises gets extracted out of it in `xlisp-test`. With a `nil` value for `*exercises*` the test suite silently succeeds.

There will be two pull requests.

The first will add code that makes `xlisp-test` fail when `*exercises*` is `nil`.

The second will correctly extract the exercise name from `config.json`.

wobh:
I guess there will just be one PR? When I updated my master branch, it updated the first PR to work.

verdammelt:
Good catch and how embarrassing.

wobh:
Pretty much my favorite kind of bug.

In summary:

https://travis-ci.org/wobh/xlisp/builds/230969489 shows that the `xlisp-test:test-exercises` fail if `*exercises*` is `nil` before the test run (note: error status was later changed to `3`).

https://travis-ci.org/wobh/xlisp/builds/231007721 shows the build passing with tests running.

PR #150, opened 2017-05-10
gh-149 1/2: fail when *exercises* is undefined

wobh:
Add a condition to warn and fail if `*exercises*` is undefined in `test-exercises`.


wobh:
^^ That's what is supposed to happen.

verdammelt:
Looks good - thanks.

Issue #138, opened 2017-02-26
Replace CIM (deprecated)


verdammelt:
CIM has been deprecated by its author (http://keens.github.io/blog/2017/01/29/deprecating_cim/) Our build uses this tool. Perhaps switch to roswell (https://gist.github.com/fukamachi/52dce85caf477043ffeb)

wobh:
I'm reviving this and will work on it this week. Roswell looks like an excellent replacement, and seems like it'll be good for TravisCI. I also think having this in place will be useful for local xlisp development, such as for #140.

verdammelt:
Thank you @wobh I haven't had time for much work on xlisp lately at all.

wobh:
A great sign! https://travis-ci.org/wobh/xlisp/builds/231021325

verdammelt:
Believe this is covered by #151  so closing.

PR #151, opened 2017-05-11
gh-138: setup travis with ros

wobh:
clasp and cmu-bin will not be supported for now


verdammelt:
This looks pretty good.

Looks like we can get rid of that helper script that installed CIM.

Also any changes to the docs needed?

wobh:
I'm working on the docs.

wobh:
I felt like I had to simplify the README docs a bit, taking out the part about how to locally test all installed lisp implementations with Roswell. I couldn't find a way to do that like we could with CIM. We might just have to advise people to set up TravisCI for their fork.

verdammelt:
The doc changes look good. But before I merge let me "test" them by following them and see how it work. Will try to do that tonight/tomorrow night.

PR #152, opened 2017-05-22
Add the file loading line necessary to install Quicklisp

Aluriak:
Without this line, lisp do not know `quicklisp-quickstart` name and raise an error.
Tested on clisp 2.49.


verdammelt:
@Aluriak thanks for catching this.

PR #170, opened 2017-08-07
Travis improvement

verdammelt:
Experimenting with using container builds with caching and checking just which implementations work and which don't.


verdammelt:
Paging @wobh... 

Thoughts?

wobh:
Quicking skimming through this, and this seems good. Should run faster on TravisCI too, yes?

Aside: I won't have much time for this track for the next couple of months. We're getting ready for a move across the country.

verdammelt:
Yes, it will usually be faster (after the initial run when there is no cache).

Good luck on the move!

PR #178, opened 2017-10-13
added hello-world exercise

serialhex:
Why this one wasn't implemented yet... nobody knows... 👻

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,6 +3,16 @@
   "active": true,
   "exercises": [
     {
+      "uuid": "d4af6cfa-d468-15b5-dcdf-7d717ddf6556",
+      "slug": "hello-world",
+      "core": true,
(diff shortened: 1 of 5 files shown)


serialhex:
...so idk if this is right, but it looks like sbcl et.al. are being built **from source** which is why the build took 8+ minutes and thus failed... Is that right?  Do the other tracks build the compilers from source every time also???  If I pushed a new Ruby exercise would I have to wait for MRuby to finish building????? 😕

verdammelt:
The building from source thing is caused by the tool we are using for integration with Travis. I am unhappy with it - but haven't had the chance to find a new thing yet.

Also we don't have a hello-world exercise on this track since the track pre-dates the idea of that exercise. But why not - let's have one!

PR #186, opened 2017-10-20
Implementing sublist

Average-user:
One more for Common Lisp


verdammelt:
Thanks! But why did you remove the 'source' section?

exercises/sublist/sublist.lisp
```diff
@@ -0,0 +1,8 @@
+(defpackage #:sublist
+  (:use #:common-lisp)
+  (:export #:sublist))
+
+(in-package #:sublist)
+
+;; what is list1 of list2 (sublist, superlist, equal or unequal)
```
verdammelt: A thought: make this comment into a doc string.

exercises/sublist/sublist-test.lisp
```diff
@@ -0,0 +1,87 @@
...
+#-xlisp-test (load "sublist")
+
+(defpackage #:sublist-test
+  (:use #:common-lisp #:lisp-unit))
+
+(in-package #:sublist-test)
+
+
+;; Equal lists tests:
+(define-test empty-equal
```
verdammelt: How about having the test names be the same as as the canonical data descriptions (in kebab-case): https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/canonical-data.json

config.json
```diff
@@ -30,6 +30,16 @@
       ]
     },
     {
+      "uuid": "428e5b28-31b6-4c84-8eae-d7753d45efc1",
+      "slug": "sublist",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+        "Strings"
```
verdammelt: Not really about strings is it?

Average-user:
@verdammelt 
>Thanks! But why did you remove the 'source' section?

Cause it was from another exercise (copy pasting problems), and [sublist](https://github.com/exercism/problem-specifications/blob/master/exercises/sublist/metadata.yml) doesn't specify a source.

verdammelt:
This exercise might also be solvable using the set functions? `intersection` `set-difference` etc. right? Is "Set" or "Sets" a keyword for topics? I can't find the list of exercise topics that had been discussed right now...

RE: removing the source: ah a dreaded copy-paste error! Squash the commits in that case.

Average-user:
>  I can't find the list of exercise topics that had been discussed right now...

I know where it is.

config.json
Average-user: Tell that to the person who named the exercise. But fair enough.

Average-user:
@verdammelt 
>Is "Set" or "Sets" a keyword for topics?

It is not about sets, they do not have structure, strings or lists are like mathematical sequences, not sets.

config.json
Average-user: I'm crazy, I've been think that the name was substrings the whole time. Sorry about that.

verdammelt:
Thanks for updating the topic. And for updating the test names..

I want to do another once-over but I think we'll merge this in later today.

Thanks!

Average-user:
It's ok, Im still thinking about another topic that fits

verdammelt:
@Average-user Can you put a link in here for the topic list?

Average-user:
[#884(https://github.com/exercism/problem-specifications/issues/884)

PR #187, opened 2017-10-22
added acronym exercise

serialhex:
You never realize how thankful you are for Regex's until you try to do some string parsing without them...

### DIFF
--- a/config.json
+++ b/config.json
@@ -8,7 +8,8 @@
       "unlocked_by": null,
       "difficulty": 1,
       "topics": [
-
+        "Strings",
+        "Transforming"
       ]
     },
     {
--- a/config.json
+++ b/config.json
@@ -2,6 +2,18 @@
   "active": true,
   "exercises": [
     {
+      "uuid": "6efa02be-0f9e-1d80-7efb-536d383535c98b1d883",
+      "slug": "acronym",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+
+      ]
+    },
+    {
+      "uuid": "d4af6cfa-d468-15b5-dcdf-7d717ddf6556",
+      "slug": "hello-world",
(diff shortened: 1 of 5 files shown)


exercises/acronym/acronym-test.lisp
```diff
@@ -0,0 +1,48 @@
+(ql:quickload "lisp-unit")
+#-xlisp-test (load "acronym")
+
+(defpackage #:acronym-test
+  (:use #:common-lisp #:lisp-unit #:acronym))
```
verdammelt: The build is failing. I think it is because you are specifing that this package uses the `acronym` package. When the tests run the packages defined by the exercises are deleted after the tests run. Some implementations are complaining that `acronym` cannot be deleted because it is in use by `acronym-test`.

config.json
```diff
@@ -3,6 +3,16 @@
   "active": true,
   "exercises": [
     {
+      "uuid": "6efa02be-0f9e-1d80-7efb-536d383535c98b1d883",
+      "slug": "acronym",
+      "core": false,
+      "unlocked_by": null,
+      "difficulty": 1,
+      "topics": [
+
```
verdammelt: We should have at least one topic. "Strings" is probably a good one.

serialhex:
Added both Strings & Transforming as that's what C# has. Also fixed a merge conflict... I think I borked the history a bit and made it not-so-nice 😿 .

verdammelt:
We can do a 'squash and merge' which will remove all history and make a single commit to master.

serialhex:
some of my commits were doubled-up, as I saw the potential merge conflict on Github and so locally I rebased, but then my remote was off and then I merged that with resulted in another conflict and blah blah blah... git user fart blah blah...

I reverted and force pushed, so this should be kosher once Travis gives its okay.

PR #188, opened 2017-10-22
the results of running `configlet generate . --spec-path ..\problem-specifications` && `configlet fmt .` on the track

serialhex:
I noticed after running `configlet generate...` while doing the acronym exercise that a *bunch* of items weren't correct according to that tool, so I reverted & made a separate pull request w/ all the changes the aforementioned commands gave.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1,428 +1,428 @@
 {
-  "language": "Common Lisp",
   "active": true,
   "exercises": [
     {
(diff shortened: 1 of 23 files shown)


verdammelt:
I wonder if there is a way to get the build to fail if this sort of thing happens again? I tried running `./bin/configlet lint .` which is what the build does: no output - successful return code.

I guess this will have to be periodic maintenance on the track...

PR #194, opened 2017-12-12
Updated installation steps for Common Lisp

defunkydrummer:
The description has been changed to suggest installing Portacle, which takes the most widely used tools for Common Lisp development (SBCL, Emacs, SLIME, ParEdit, and Quicklisp) and "wraps them up" in a pre-configured environment that is easy to install. This is particularly suited to beginners, and popular within the community, so it should be included.

### DIFF
--- a/docs/INSTALLATION.md
+++ b/docs/INSTALLATION.md
@@ -1,3 +1,29 @@
+*What's needed*
+
+The basic items needed for developing in Common Lisp are:
+
+- A Lisp "Implementation", which will allow you to compile and execute code, as well as supply a REPL.
+- A Text editor with facilities for Lisp code. 
+- (Nice to have) The often-used reference for the Common Lisp language, the "Common Lisp Hyperspec", available online [here](http://www.lispworks.com/documentation/HyperSpec/Front/Contents.htm)
+
+*Fast Start*
+
+Newbies to CL, particularly ones with previous experience with Emacs, can get an easy & quick start by installing Portacle, the Portable Common Lisp Environment. Free and full featured, it works in all common platforms and comes pre-configured "out of the box". This will supply:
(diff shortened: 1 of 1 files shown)


docs/INSTALLATION.md
```diff
@@ -1,3 +1,29 @@
...
+
+- A Lisp implementation: SBCL, including Quicklisp and ASDF.
+- A Text editor: Emacs
+... with lots of add-ons for writing with Lisp code: 
+- SLIME, the "Superior Lisp Interaction Mode for Emacs" turns Emacs into a Common Lisp IDE
+- ParEdit, which makes working with parentheses easy. 
+
+Everything comes already configured out of the box. Install Portacle by downloading from the [front page](https://portacle.github.io/)
+
+Note: Emacs (text editor) can be disorienting at first, if you're not accustomed to it. Fortunately there are many primers on Emacs and SLIME available online.
```
verdammelt: I think we need a link or two here. For these primers.

defunkydrummer: If you want i can expand that very same page with about 40 lines of text that would tell you the essentials for Emacs. Otherwise, i'll look for a tutorial, however sometimes the tutorials are for using Emacs in general, and for Lisp coding you need to know just a tiny subset of what Emacs can do.

verdammelt: I think a link to an Emacs primer and a SLIME primer might be sufficient. If we can't find them (reasonably) quickly then let's just merge this PR - it improves the documentation.

verdammelt:
@defunkydrummer I'm going to merge this in; but I still think having some links in the section mentioend above would be a good idea. Please open a new PR if you find any you think would be good.

PR #197, opened 2017-12-31
Fix invalid uuid

sjwarner-bp:
Configlet previously could generate invalid uuids.

It looks like `acronym` had an invalid uuid.


verdammelt:
Thanks for catching this!

PR #199, opened 2018-01-15
Move hello-world exercise to the beginning of the track

benreyn:
It appears that the hello-world exercise is not the first on in the track
for common-lisp. This PR should fix that.


verdammelt:
Thanks.

PR #204, opened 2018-02-19
Acronym: remove HTML test case

daveyarwood:
My usual approach to solving this exercise (split on hyphen or space, take all first characters and uppercase them) failed on the HTML example, which made me curious, so I dug a little deeper and found that the test case has been [removed from the canonical data](https://github.com/exercism/problem-specifications/pull/788).

### DIFF
--- a/exercises/acronym/acronym-test.lisp
+++ b/exercises/acronym/acronym-test.lisp
@@ -22,11 +22,6 @@
     "ROR"
     (acronym:acronym "Ruby on Rails")))
 
-(define-test html-test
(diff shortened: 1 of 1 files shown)


verdammelt:
That HTML case was a tough one. If it has been removed from the canonical data then removing from this track is the right thing to do.  Thanks.

(I'm just checking on why the build failed and then will merge)

verdammelt:
The build is failing because the tool roswell (which we use to test on multiple lisps) is having trouble installing clisp.

PR #205, opened 2018-02-19
Fix the getting started/help guide links to be correct

azrazalea:
Guess: At one point this track was called lisp not common-lisp.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ that fail tests
 ### Development setup
 
 Direct contributions to the Common Lisp code in xLisp are always welcome.
-Refer to [the Getting Started Guide](http://exercism.io/languages/lisp) for the
+Refer to [the Getting Started Guide](http://exercism.io/languages/common-lisp) for the
 Common Lisp track to get your environment set up if needed.
 
 New exercises or changes to existing ones can be submitted via a pull
@@ -104,4 +104,3 @@ error code if there are problems):
 The example implementation should use nothing outside of the Common
 Lisp specification. It should be well designed and not too clever.
 (Save the clever implementation for your submission on Exercism.io.)
-
--- a/config/exercise_readme.go.tmpl
+++ b/config/exercise_readme.go.tmpl
@@ -6,7 +6,7 @@
 {{ end }}
 ## Setup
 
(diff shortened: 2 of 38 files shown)


azrazalea:
The travis build failure is due to unrelated issues with clisp. I will take a look at fixing them if I have time.

azrazalea:
Triggering rebuild via close/open PR since it looks like this may just be a temporary error with roswell.

azrazalea:
Running again did not help and roswell installs clisp fine locally using what looks like the same steps. Not sure if that `.tar.gz` is getting deleted after being downloaded or if something else is going on.

verdammelt:
RE: your guess: quite right it was. Renamed it to be more specific. But not after lots of links etc. were embedded everywhere. Thank you for this PR.

The roswell failure is annoying but obviously not caused by this PR.

azrazalea:
@verdammelt I did not! Trying to do so to verify the results are the same, and I found https://github.com/exercism/x-api/blob/master/CONTRIBUTING.md which appears to be missing the section on configlet (found that link via the README.md in this repository). 

I played around with it a bit and not getting `configlet generate` to work right. Can you point me to the docs?

verdammelt:
re: point you to the docs - I don't know if I can! :/

How about I merge this PR later (tonight) and then we'll do a little clean-up on any remaining things.

verdammelt:
Maybe this doc can help: https://github.com/exercism/configlet/blob/master/README.md

azrazalea:
With the help of that document I:

1. Cloned the `problem-specifications` to be a sibling of the `common-lisp` folder
2. With my pwd being the `common-lisp` folder I ran:
    ```
    bin/configlet generate . --spec-path ../problem-specifications
    ```

With this, a couple changes were made. One removed an actual line from a README.md, the other added trailing whitespace (which my editor auto removes so may have been due to my changes).

Comitting these for your perusal @verdammelt

verdammelt:
Thank you. I think it is fishy that the bob exercise was changed slightly - I'll investigate after merging this. So I'm going to merge - but might have a bit more conversation here if I need your help to figure out what happened. OK?

verdammelt:
Well - everything looks good. Thanks!

PR #209, opened 2018-06-13
fix link to successful lisp

sjwarner-bp:
Looks like while David Lamkins is reworking their website, they've recommended a new link to find out about the Successful Lisp book.

I've just amended the RESOURCES.md accordingly 🙂


benreyn:
Thanks @sjwarner-bp !

PR #219, opened 2018-08-25
Change sublist return type to keywords

kephas:
Returning keywords makes it possible to use CASE.


NobbZ:
You need to fix the example as well.

NobbZ:
> as the canonical data exepcts strings

Just because there are no atoms or enumerations in JSON, and a string therefore is the most agnostic way to represent those.

In erlang I represent them as atoms, and it will be easy to continue to do so, even after creating an generator. All I need to do, is to tell the generator to create the code that creates the atom with the value `sublist` instead of the string `"sublist"`. 

In about any language that has atoms, their usage would be prefered over the usage of strings in idiomatic code in similar situations, as especially comparing them is usually much more efficient than comparing strings.

verdammelt:
That is a good argument. I'll merge this in.

@kephas / @NobbZ either of you interested in creating PRs for other exercisese that can be improved in this way?

kephas:
I'll do it if I encounter one. Don't hesitate to ping me if you see one.

PR #225, opened 2018-10-20
Implement octal exercise

serialhex:
BAM! Just like the title says...


serialhex:
No problem, I'll just go implement that instead.

PR #226, opened 2018-10-20
Implement luhn exercise

serialhex:
SHAZAM!!!

Another one!


serialhex:
BAH!!!! Forgot some bookeeping stuff.  I'll fix and update my PR in a minute, for both this and the Octal exercise.

exercises/luhn/example.lisp
```diff
@@ -0,0 +1,36 @@
+(in-package #:cl-user)
+(defpackage #:luhn
+  (:use #:cl)
+  (:export #:is-valid))
+
+(in-package #:luhn)
+
+(defun is-valid (input)
+  (let ((num (remove #\Space input)))
+    (if (or (< (length num) 2)
```
verdammelt: Might I suggest turning this `if` into an `unless`? Or reverse the predicate and use a `when`?

exercises/luhn/example.lisp
```diff
@@ -0,0 +1,36 @@
...
+                            (let ((num (digit-char-p (car lst))))
+                              (cond
+                                ((= 1 (mod elm 2))  num)
+                                ((< num 5)          (* 2 num))
+                                (t                  (- (* 2 num) 9)))))
+                         (1+ elm))
+                  (= 0 (mod tot 10)))))
+    (rec-num (nreverse
+              (map 'list
+                   (lambda (x) x)
```
verdammelt: `identity`? (http://l1sp.org/cl/identity)

verdammelt: And in this case - is this just meant to be a `coerce`?

exercises/luhn/example.lisp
```diff
@@ -0,0 +1,36 @@
...
+(defpackage #:luhn
+  (:use #:cl)
+  (:export #:is-valid))
+
+(in-package #:luhn)
+
+(defun is-valid (input)
+  (let ((num (remove #\Space input)))
+    (if (or (< (length num) 2)
+            (not (reduce (lambda (t/f char)
```
verdammelt: Is this checking that every character is a digit character? Perhaps `every` along with `digit-char-p`?

wobh:
The example doesn't have to be an exemplar, but it should be the "simplest" solution that can "simply" run in the test suite. Name all relevant parts, use standard tools and methods, and yet economize forms. I get something like this:

```lisp
(defun valid-p (input)
  (when (and (< 1 (length input))
             (every (lambda (char)
                      (or (digit-char-p char)
                          (char= #\Space char)))
                    input))
    (flet ((double-but-nine (n)
             (let ((nn (* 2 n)))
               (if (< 9 nn) (- nn 9) nn))))
      (loop
         with checksum = 0
         with digits = (nreverse
                        (remove-if #'null
                                   (map 'list
                                        #'digit-char-p
                                        input)))
         for index = 1 then (1+ index)
         for digit in digits
         do (incf checksum
                  (if (evenp index)
                      (double-but-nine digit)
                      digit))
         finally
           (return (zerop (mod checksum 10)))))))

But another solution might occur to you.

(assert (valid-p "4539 1488 0343 6467"))
(assert (not (valid-p "8273 1232 7352 0569")))
```

serialhex:
@verdammelt instead of commenting one-by-one, I'll just say it here: I was doing the derp, thanks for the double-check.  I don't even like `(if X nil Y)` type stuff... I must have been really out of it!

Made the changes and just pushed.  The tests should run in a minute and be all kosher!

verdammelt:
Let's merge this!

PR #233, opened 2018-11-06
Add instructions for adding new exercise.

verdammelt:


docs/template-test.lisp
```diff
@@ -0,0 +1,17 @@
...
+
+;;;
+;;; tests go here...
+;;; /e.g./ (define-test a-test (assert-equal 1 2))
+;;;
+
+#-xlisp-test
+(let ((*print-errors* t)
+      (*print-failures* t))
+  (run-tests :all :twelve-days-test))
```
wobh: s/twelve-days/exercise/

verdammelt: drat! good catch - thanks.

README.md
```diff
@@ -73,14 +73,38 @@ running `ros list installed`.
...
+   <path-to-problem-specifications>`.
+3. Using the canonical data found in the `problems-specifications`
+   repository create tests for the exercise. A template for the test
+   file can be found in
+   [`docs/template-test.lisp`](docs/template-test.lisp) (change
+   `exercise` to the name of your exercise in kebab-case).
+4. Create an example implementation which passes the tests. This does
+   not need to be clever or amazing (save that for your own solution
+   submission), but it should have good style and idiom and use
+   nothing outside of the Common Lisp specification.
```
wobh: I think we should make this last statement stronger. For example, in any exercise where they say you shouldn't use language provided solutions (which I think is silly, but whatevs) the examples should *totally* use language provided solutions. The base conversion and sequence functions exercises seem to emphasize that. Examples are tests of the product, the product is tests.

verdammelt: I am mixed about exercises with the "don't use built-in" caveat. I think they should be solved twice - once without built in features and once with (or _vice_versa_). So the mentee understands what the built-in feature is providing for them.

As for this above statement perhaps something like:

> ...(and if the exercise suggests not to use built-in features, the example code should ideally *use* those built-in features to highlight the functionality).

PR #252, opened 2019-07-15
Rename square-of-sums to square-of-sum to match the tests

sys-fs:
The tests use `square-of-sum`, but the `squares` package exports `square-of-sums`. This fixes the typo.


verdammelt:
Thanks for catching this!

PR #253, opened 2019-07-29
Add test case for empty string input

OrdoFlammae:
Title says it all. I think the program should also check for an empty string, since that's another input that probably would be seen as invalid.

### DIFF
--- a/exercises/two-fer/example.lisp
+++ b/exercises/two-fer/example.lisp
@@ -5,4 +5,4 @@
 (in-package #:two-fer)
 
 (defun twofer (name)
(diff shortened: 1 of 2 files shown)


exercises/two-fer/example.lisp
```diff
@@ -5,4 +5,4 @@
 (in-package #:two-fer)
 
 (defun twofer (name)
-  (format nil "One for ~a, one for me." (or name "you")))
+  (format nil "One for ~a, one for me." (if (or (null name) (equal name "")) "you" name)))
```
verdammelt: We should use `string=` here. Generally the "rule" is to use the most specific equality operator possible.

PR #277, opened 2020-02-02
Fix embarrassing loading of from file

### LINKED ISSUE #276: Exercise Bob: Loads incorrect file
[`bob-test.lisp`](https://github.com/exercism/common-lisp/blob/master/exercises/bob/bob-test.lisp) line 6 tries to load `example.lisp`, which, along with being a completed solution, is not in the downloaded files for the exercise. Line 5 should be uncommented, line 6 should be removed or commented out.

verdammelt:
Fixes #276

### DIFF
--- a/exercises/bob/bob-test.lisp
+++ b/exercises/bob/bob-test.lisp
@@ -2,8 +2,7 @@
 ;;; bob v1.6.0
 ;;;
 (ql:quickload "lisp-unit")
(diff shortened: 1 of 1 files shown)


verdammelt:
Overriding the failing build as it is failing due to the Roswell tool - not the code change - and as we have found out the code change would not be caught in the build anyway.

PR #284, opened 2020-06-24
replaced `is-valid` with `validp` in exercise `luhn`

chuchana:


TheLostLambda:
Hi @chuchana ! This is a good catch! Using the predicate naming seems much more idiomatic! Do you prefer `valid-p` over `vaildp`?

TheLostLambda:
Additionally, I think we will need to update the naming in `exercises/luhn/example.lisp` to get the Travis build passing :)

chuchana:
Hi, thanks, but I didn't catch it, the mentor @verdammelt did.

I find `valid-p` more readable, though I think the convention is to just append a `p` for short/single words. Should I change it to `validp`?

TheLostLambda:
I'm not too fussed about `valid-p` vs `validp`, but glancing through the standard, it seems like `validp` is more idiomatic here. I might lean towards `validp` for that reason, but I'm alright with merging either :) Just let me know!

TheLostLambda:
Looks good to me!

verdammelt:
@chuchana thanks for contributing to the track!

@TheLostLambda thanks for handling the merging.

PR #286, opened 2020-10-04
Port "Binary Search" exercise

wsgac:


wsgac:
Hey there. Just a reiteration of my question from another PR I had to cancel... Are you guys planning to participate in this year's Hacktoberfest?

verdammelt:
Thanks for submitting this PR. It generally looks good but I have two questions comments I'd like us to discuss before merging:

1. Might it be more idiomatic to evalute to multiple values (like `gethash`) rather than signal and error?
2. Since we are not (I do not think) that *binary search* is being performed - do we allow students to implement their solution with `find`? Perhaps we need explicit instructions *not* to use `find` (or similar) for this exercise?

verdammelt:
> ad 1. Yep, raising exceptions was probably not the lispiest choice out
> there. Actually, I don't think we need multiple values. An integer or
> `NIL` will do to distinguish the item's nonexistence.

I'd go for `nil` myself, or go the `gethash` direction and using
multiple values: value (may be `nil`) & value-found-p.

> ad 2. I agree. I can make a note to that effect.

Thanks.

verdammelt:
Thanks for writing this exercise! I'll mege it and then i'll have to go do it so I can mentor it :)

PR #287, opened 2020-10-05
Stub functions to help users with exercises

dnmfarrell:
These help users complete the exercise as they know what interface they
need to provide in order to run the test script.

### DIFF
--- a/exercises/acronym/acronym.lisp
+++ b/exercises/acronym/acronym.lisp
@@ -4,3 +4,7 @@
   (:export #:acronym))
 
 (in-package #:acronym)
+
+(defun acronym (str)
+  "Returns the acronym for a noun of tech jargon."
+  )
--- a/exercises/allergies/allergies.lisp
+++ b/exercises/allergies/allergies.lisp
@@ -5,3 +5,11 @@
   (:export #:allergic-to-p #:list))
 
 (in-package #:allergies)
+
+(defun allergic-to-p (score allergen)
+  "Returns true if given allergy score includes given allergen."
+  )
+
+(defun list (score)
+  "Returns a list of allergens for a given allergy score."
+  )
--- a/exercises/anagram/anagram.lisp
+++ b/exercises/anagram/anagram.lisp
@@ -4,3 +4,7 @@
   (:export #:anagrams-for))
 
 (in-package #:anagram)
+
+(defun anagrams-for (subject candidates)
+  "Returns a sublist of candidates which are anagrams of the subject."
+  )
--- a/exercises/beer-song/beer-song.lisp
+++ b/exercises/beer-song/beer-song.lisp
@@ -4,3 +4,10 @@
   (:export #:verse #:sing))
 
 (in-package #:beer)
+(defun verse (n)
+  "Returns a string verse for a given number."
+  )
+
+(defun sing (start &optional (end 0))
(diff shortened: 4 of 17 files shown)


verdammelt:
A conversation the mentors and maintainers had (yes more than just me) was about just such stub functions. At the time (1 or 2 years ago) we choose not to have stubs except in the first few exercises as we felt that the tests & documentaiton are what should clarify the interface and after the first few exercises the student knows how to write a function.

Also note that your stub functions could cloud the waters. They are all functions which ignore all their arguments (if they have any) and evalute to a string. I'm assuming that was not your intent? Do the stubbed functions make the failing tests' errors clearer?

dnmfarrell:
Ah, I thought they were (unintentionally) missing.

As a user, the experience of discovering the required interface by running the tests to discover the missing function names, then multiple times to figure out how many args to accept isn't great. If I open the test file, I've now seen the test cases which can influence my thinking of how to code the solution.

I like the idea of describing the interface in the documentation leaving the developer free to implement the function interface as they like (required params, default values etc). Of the 9 exercises I've completed, I don't think any had the interface described in the docs though; is that work tbc?

verdammelt:
> ...I've now seen the test cases which can influence my thinking of how
> to code the solution.

Interesting I've never though of the test cases as some sort of secret.

Would your opinion be different if, like other tracks, some sort of
skipping/commenting-out is done on all but the first test and it is the
student's job to un-skip/un-comment each test in turn to simulate the
TDD cycle?

> I like the idea of describing the interface in the documentation
> leaving the developer free to implement the function interface as they
> like (required params, default values etc).

I can see your point. But the stubs need to be more than: `(defun foo
(args) "doc string")` right? That stub shows a function evaluating to a
string. It would need to at least be `(defun foo (args) "doc string"
nil)`, that will be less problem in the cases were `nil` is not a valid
value (this being Common Lisp - `nil` is *always* a valid response it
seems :) ).

> Of the 9 exercises I've completed, I don't think any had the interface
> described in the docs though; is that work tbc?

The official description of the tests and documents is in the
`problem-specifications` repository. Updating these tests from that is a
manual process however and I have not done it for a while.

dnmfarrell:
Maybe it's just my weird way of thinking about it. 

RE: un-commenting the tests; I guess that helps a little with my complaint as to not seeing all the test cases the first time I run the file.

I see what you're getting at about the stub entry, I agree it would be better to let the developer write it. Perhaps instead it could be documented like this?

```lisp
(in-package #:cl-user)
(defpackage #:beer
  (:use #:common-lisp)
  (:export #:verse #:sing))

;;; verse - accepts an integer and returns a string of the verse
;;; for the integer

;;; sing - accepts a starting integer and optionally an ending
;;; integer returning a string of all the verses from the start to
;;; to the end, or 0 if no ending integer was provided
```

Then the developer is clear what they need to implement, what values the functions may receive and what return types the functions have to provide, but they can still define the function signature as they see fit, give the parameters whatever names they deem appropriate, etc.

verdammelt:
After letting this rest in my head for a while and reviewing what I'm doing on the v3 version of this track (which has stub functions) I think having stub functions is fine by me. So I'm thinking of merging this soon.

verdammelt (approved):
Deciding to approve this to get the work on the main branch before too much V2<->V3 practice exercise shenanigans happen.

PR #298, opened 2021-01-20
Prepare for v3 by removing usage of sharpsign-dot in example source file.

verdammelt:
The v3 representer will not (and can not) handle #'.. Since I plan on
testing the representer by running it against all exercise examples -
so this needs to be fixed.

### DIFF
--- a/exercises/binary-search/.meta/tests.toml
+++ b/exercises/binary-search/.meta/tests.toml
@@ -1,0 +1,34 @@
+[canonical-tests]
+
+# finds a value in an array with one element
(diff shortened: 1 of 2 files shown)


TheLostLambda (approved):
Not sure if I'm actually a "Reviewer with write access" to this repo, but if so, this looks good to me! Hopefully we'll eventually be able to hack together something for reading `#` in the v3 tooling, but this is a good stop-gap

verdammelt:
@TheLostLambda on the question if you are a reviewer with write access - we'll have to figure that out for post Feb 1 when this becomes *the* repo for the track.

PR #338, opened 2021-01-29
Add Track Tags

### LINKED ISSUE #326: [v3] Add tags
_This issue is part of the migration to v3. You can read full details about the various changes [here](https://github.com/exercism/v3-launch/)._

In Exercism v3, tracks can be annotated with tags. This allows searching for tracks with a certain tag combination, making it easy for students to find an interesting track to join.

Tags are specified in the top-level `"tags"` field in the track's `config.json` file and are defined as an array of strings, as specified in [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#tags).

## Goal

The `"tags"` field in the `config.json` file should be updated to contain the tags that are relevant to this track. The list of tags that can be used is listed in [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#tags).

## Example

```json
{
  "tags": [
    "runtime/jvm",
    "platform/windows",
    "platform/linux",
    "paradigm/declarative",
    "paradigm/functional",
    "paradigm/object_oriented"
  ]
}
```

## Tracking

https://github.com/exercism/v3-launch/issues/1

TheLostLambda:
This should close #326

Let me know if you disagree with any of the tags! This was just my first swing at things!

### DIFF
--- a/config.json
+++ b/config.json
@@ -1103,5 +1103,22 @@
     }
   ],
   "key_features": [],
-  "tags": []
+  "tags": [
+    "paradigm/functional",
+    "paradigm/imperative",
+    "paradigm/object_oriented",
+    "paradigm/procedural",
+    "typing/dynamic",
+    "typing/weak",
+    "execution_mode/compiled",
+    "execution_mode/interpreted",
+    "platform/windows",
+    "platform/mac",
+    "platform/linux",
+    "runtime/language_specific",
+    "used_for/artificial_intelligence",
+    "used_for/backends",
+    "used_for/games",
+    "used_for/scripts"
+  ]
 }


config.json
```diff
@@ -1103,5 +1103,22 @@
...
   ],
   "key_features": [],
-  "tags": []
+  "tags": [
+    "paradigm/functional",
+    "paradigm/imperative",
+    "paradigm/object_oriented",
+    "paradigm/procedural",
+    "typing/dynamic",
+    "typing/weak",
```
verdammelt, line 1112: Isn't it strongly typed?

TheLostLambda: Right you are! I think I misinterpreted the meaning of weak vs strong! Sorry about that!

verdammelt: I had to look it up myself :/

iHiD: I've just read the Wikipedia page and I literally still have no idea what they mean 🤷

verdammelt: C is weakly typed - pointers can be interpreted as integers or strings etc.
JavaScript is weakly typed - strings are numbers or numbers are strings in some cases.

Common Lisp has a type for every value and those types will not change during the scope of that value. Even if you use `coerce` you are not changing the type of the value - but are producing a new value with a new type.

Thus I'd say Common Lisp is strongly typed.

iHiD: Merge it! :)

verdammelt (approved):

Issue #340, opened 2021-01-30
The old (removed) CI system needs to be replaced.


verdammelt:
Previously the CI system included running every exercise test with the exemplar code to ensure that at least worked. However with the V3 merge and a coincidental failure of Roswell (bad timing) we removed that portion of CI. We will need to add something back in to do this.

verdammelt:
Merged in the start of the Test Exercise CI system. We can improve it as we go.

PR #344, opened 2021-01-30
Stub CI for testing all exercises.

verdammelt:
This PR only contains changes need to be able to `asdf:test-system` a system which *could* test all exercises. ~~However it does not do that presently.~~ And it actually runs the tests!

I am creating this PR now because I think that the pattern here may be useful for other CI tasks and perhaps this part of the process could be expanded and improved separate from the actual testing of exercises. (_e.g._ multiple platforms could be tested, multiple implementations could be tested or caching could be added to speed up teh runs.)

(_cf._ #340)


verdammelt:
I think this is a workable system. It runs all exercises (both fiveam & lisp-unit based) and reports the results. If any failures/errors/etc were found the build fails.

Example failure: https://github.com/exercism/common-lisp/runs/1799949591?check_suite_focus=true

This showed a bunch of our exercises (including a new v3 concept exercise) which had problems which were then fixed and the test ran green: https://github.com/exercism/common-lisp/runs/1799954632?check_suite_focus=true

TheLostLambda (approved):
This is a super impressive amount of work, and looks good to me! It's also nice to have so many of the exercises polished up and passing tests!

Thanks a ton for this!

PR #349, opened 2021-02-07
Add key features for this track.

### LINKED ISSUE #325: [v3] Add key features
_This issue is part of the migration to v3. You can read full details about the various changes [here](https://github.com/exercism/v3-launch/)._

In Exercism v3, each track must specify exactly six "key features". Exercism uses these features to highlight the most interesting, unique or "best" features of a language to a student.

Key features are specified in the top-level `"key_features"` field in the track's `config.json` file and are defined as an array of objects, as specified in [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#key-features).

## Goal

The `"key_features"` field in the `config.json` file should be updated to describe the six "key features" of this track. See [the spec](https://github.com/exercism/v3-docs/blob/main/anatomy/tracks/config-json.md#key-features).

## Example

```json
{
  "key_features": [
    {
      "icon": "features-oop",
      "title": "Modern",
      "content": "C# is a modern, fast-evolving language."
    },
    {
      "icon": "features-strongly-typed",
      "title": "Cross-platform",
      "content": "C# runs on almost any platform and chipset."
    },
    {
      "icon": "features-functional",
      "title": "Multi-paradigm",
      "content": "C# is primarily an object-oriented language, but also has lots of functional features."
    },
    {
      "icon": "features-lazy",
      "title": "General purpose",
      "content": "C# can be used for a wide variety of workloads, like websites, console applications, and even games."
    },
    {
      "icon": "features-declarative",
      "title": "Tooling",
      "content": "C# has excellent tooling, with linting and advanced refactoring options built-in."
    },
    {
      "icon": "features-generic",
      "title": "Documentation",
      "content": "Documentation is excellent and exhaustive, making it easy to get started with C#."
    }
  ]
}
```

## Tracking

https://github.com/exercism/v3-launch/issues/5

verdammelt:
I tried to pick what I think are the important features for Common Lisp. I'm sure there will be disagreement, and I am open to discussion.

Also I'm sure some of my wording could be improved. The 100 character limit was challenging.

Fixes #325.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1197,7 +1197,7 @@
     {
       "icon": "feature-inspectable",
       "title": "Inspectible",
-      "content": "Common Lisp is fully inspectable with functions such as `inspect`, `documentation`, and `describe`."
+      "content": "Common Lisp is fully inspectable with built-in functions for exploring every part of the language."
     },
     {
       "icon": "feature-homoiconicity",
@@ -1207,17 +1207,17 @@
     {
       "icon": "feature-extensible",
       "title": "Extensible",
-      "content": "Generic functions allow for extending existing features by specializing, overriding or wrapping."
+      "content": "Generic functions, reader-macros, and flexible namespacing allow for the easy creation of DSLs"
     },
     {
       "icon": "feature-standardized",
       "title": "Standardized",
-      "content": "The language is very stable, code written 30+ years ago will still run 30+ years from now."
+      "content": "The language is very stable, with multiple implementations to match your needs – JVM, embedded, etc."
     },
     {
-      "icon": "feature-multiple-implementations",
-      "title": "Multiple Implementations",
-      "content": "Multiple implementations provide choices to match your needs: JVM? embedded? optimized?"
+      "icon": "feature-multiparadigm",
+      "title": "Multiparadigm",
+      "content": "Whether you prefer a functional style or something more object-oriented, Lisp adapts to your needs."
     }
   ],
   "tags": [
--- a/config.json
+++ b/config.json
@@ -1188,7 +1188,38 @@
       "blurb": "TODO: add blurb for scope concept"
     }
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      "icon": "feature-interactive",
+      "title": "Interactive",
+      "content": "Common Lisp has a full featured REPL, development can happen as a 'conversation' with the system."
+    },
+    {
+      "icon": "feature-inspectable",
+      "title": "Inspectible",
+      "content": "Common Lisp is fully inspectable with functions such as `inspect`, `documentation`, and `describe`."
+    },
+    {
+      "icon": "feature-homoiconicity",
+      "title": "Homoiconicity",
+      "content": "Code has the same structure as data, allowing for powerful macros: code that writes code."
+    },
+    {
+      "icon": "feature-extensible",
+      "title": "Extensible",
+      "content": "Generic functions allow for extending existing features by specializing, overriding or wrapping."
+    },
+    {
+      "icon": "feature-standardized",
+      "title": "Standardized",
+      "content": "The language is very stable, code written 30+ years ago will still run 30+ years from now."
+    },
+    {
+      "icon": "feature-multiple-implementations",
+      "title": "Multiple Implementations",
+      "content": "Multiple implementations provide choices to match your needs: JVM? embedded? optimized?"
+    }
+  ],
   "tags": [
     "paradigm/functional",
     "paradigm/imperative",
--- a/config.json
+++ b/config.json
@@ -1207,7 +1207,7 @@
     {
       "icon": "feature-extensible",
       "title": "Extensible",
-      "content": "Generic functions, reader-macros, and flexible namespacing allow for the easy creation of DSLs"
+      "content": "Generic functions, reader-macros, and flexible namespacing allow for the extension of the language"
     },
     {
       "icon": "feature-standardized",


config.json
```diff
@@ -1188,7 +1188,38 @@
       "blurb": "TODO: add blurb for scope concept"
     }
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      "icon": "feature-interactive",
```
TheLostLambda, line 1193: Focusing on the features first (leaving wording, if it does need to be tweaked, until later), I think Interactive is a fantastic one! I've still never encountered another programming language that a left the full power of the language available at run-time.

An excellent choice!

config.json
```diff
@@ -1188,7 +1188,38 @@
...
+      "title": "Interactive",
+      "content": "Common Lisp has a full featured REPL, development can happen as a 'conversation' with the system."
+    },
+    {
+      "icon": "feature-inspectable",
+      "title": "Inspectible",
+      "content": "Common Lisp is fully inspectable with functions such as `inspect`, `documentation`, and `describe`."
+    },
+    {
+      "icon": "feature-homoiconicity",
```
TheLostLambda, line 1203: The classic, Lisp-unique feature! A must have!

config.json
```diff
@@ -1188,7 +1188,38 @@
...
+      "title": "Extensible",
+      "content": "Generic functions allow for extending existing features by specializing, overriding or wrapping."
+    },
+    {
+      "icon": "feature-standardized",
+      "title": "Standardized",
+      "content": "The language is very stable, code written 30+ years ago will still run 30+ years from now."
+    },
+    {
+      "icon": "feature-multiple-implementations",
```
TheLostLambda, line 1218: This is a good one to have, though I think it heavily overlaps with what I think of as "standardized". Maybe there is a way to combine the two? Or the standard is implied by this?

config.json
```diff
@@ -1188,7 +1188,38 @@
...
+      "title": "Inspectible",
+      "content": "Common Lisp is fully inspectable with functions such as `inspect`, `documentation`, and `describe`."
+    },
+    {
+      "icon": "feature-homoiconicity",
+      "title": "Homoiconicity",
+      "content": "Code has the same structure as data, allowing for powerful macros: code that writes code."
+    },
+    {
+      "icon": "feature-extensible",
```
TheLostLambda, line 1208: This is good, though I might personally place more of the focus on flexibility as well. Maybe in the description then, we could emphasise there are lots of bits of lisp that make writing DSLs and extending the language as easy as possible.

config.json
```diff
@@ -1188,7 +1188,38 @@
...
+      "title": "Extensible",
+      "content": "Generic functions allow for extending existing features by specializing, overriding or wrapping."
+    },
+    {
+      "icon": "feature-standardized",
```
TheLostLambda, line 1213: I'm certainly not opposed to this feature, but I think it could be swapped to "multi-paradigm" or "unopinionated"

config.json
```diff
@@ -1188,7 +1188,38 @@
...
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      "icon": "feature-interactive",
+      "title": "Interactive",
+      "content": "Common Lisp has a full featured REPL, development can happen as a 'conversation' with the system."
+    },
+    {
+      "icon": "feature-inspectable",
```
TheLostLambda, line 1198: I think that I really like this one as well! How do you feel about something like "self-documenting", kind of like how Emacs describes itself! I do worry it's a tad similar to interactive, but I think it's unique enough.

config.json
TheLostLambda: Maybe the feature itself could actually be flexibility and the description talking about how everything in the language is available to programmer at read, run, and compile time. Similar to how even "private" symbols can be hooked into, so you can extend a library without needing to fork it (something I really miss in a language like Rust, where half of the things I want to do **mandate** a fork / upstream work.)

TheLostLambda (commented):
I think that this is a great list to start with and that, honestly, we should probably only iterate a couple of times before merging (or we might still be doing this next year).

Some additional features that potentially deserve a spot here:

- Mult-paradigm (CL seems to allow just about anything, with excellent OOP, imperative, and functional style support!)
- I'm not sure where it belongs, maybe in with interactive or inspectable, but SLIME / SLY are probably worth some sort of mention, as that tooling is excellent and a big part of the CL experience

Finally, some resources I found for those after me:

- https://common-lisp.net/features
- http://random-state.net/features-of-common-lisp.html
- https://www.quora.com/What-is-so-special-about-LISP
- https://softwareengineering.stackexchange.com/questions/210274/does-lisp-still-have-any-special-feature-which-has-not-been-adopted-by-other-pro

Let me know what you think of my opinion-filled waffle!

verdammelt:
Oh, thanks for the link to common-lisp.net - i never remember that - and it even has the required *6* features!! :)

config.json
verdammelt: I think this is a good idea. Can you state this in 100 characters or less?

config.json
verdammelt: I think you may have a point - standard & multiple-implementation do overlap. So if we can get them into a single feature then we can add your 'multi-paradigm' feature (which i think is a good one).

config.json
verdammelt: self-documentating could work... if you think that is a better name than inspectable I'll change it. Do you think the wording should change much?

TheLostLambda:
@verdammelt I've pushed a couple of tweaks, but I'm not sure if I'm 100% sold on my own additions just yet, so let me know what you think!

config.json
```diff
@@ -1197,7 +1197,7 @@
     {
       "icon": "feature-inspectable",
       "title": "Inspectible",
-      "content": "Common Lisp is fully inspectable with functions such as `inspect`, `documentation`, and `describe`."
+      "content": "Common Lisp is fully inspectable with built-in functions for exploring every part of the language."
```
verdammelt, line 1200: 👍

config.json
```diff
@@ -1207,17 +1207,17 @@
     {
       "icon": "feature-extensible",
       "title": "Extensible",
-      "content": "Generic functions allow for extending existing features by specializing, overriding or wrapping."
+      "content": "Generic functions, reader-macros, and flexible namespacing allow for the easy creation of DSLs"
```
verdammelt, line 1210: I like the addition of reader-macros and namespacing - but I think focuing on DSL is just one outcome.

Maybe a more general outcome is that it lets you build the language up the domain of the problem rather than moving the domain down to the language.

config.json
```diff
@@ -1207,17 +1207,17 @@
...
     },
     {
       "icon": "feature-standardized",
       "title": "Standardized",
-      "content": "The language is very stable, code written 30+ years ago will still run 30+ years from now."
+      "content": "The language is very stable, with multiple implementations to match your needs – JVM, embedded, etc."
```
verdammelt, line 1215: 👍

config.json
```diff
@@ -1207,17 +1207,17 @@
...
     },
     {
-      "icon": "feature-multiple-implementations",
-      "title": "Multiple Implementations",
-      "content": "Multiple implementations provide choices to match your needs: JVM? embedded? optimized?"
+      "icon": "feature-multiparadigm",
```
verdammelt, line 1218: 👍

config.json
TheLostLambda: I was thinking a similar thing, but was struggling to get it in the word limit! I'll give that another go on Monday hopefully!

verdammelt: I think the rest of them are good to go - but this one would be good to try to improve.

I'd be happy with 'DSL' if we can't do something better.

You'll try again on Monday - and I'll try to work on this through the weekend. Early next week we'll draw a line in the sand and merge no matter what - sound good?

TheLostLambda:
@verdammelt What do you think about that wording? Is that a bit better?

verdammelt:
I like your new wording on extensible.

But by your comment - you think instead of extensible to have multi-paradigm or unopinionated? I think extensible is different than either of the latter. But perhaps I am not understanding something.

TheLostLambda:
Ah! No, sorry! That's me being a bit confusing, I think! Extensible is great on it's own and I agree it's quite different from the others! I think the comment you are referring to is the one I wrote on `standardised`, but I actually decided to keep that one and swap `multiple-implementations` for `multiparadigm`.

I'm quite happy to merge things in their current state if you are!

verdammelt:
ah, thanks for clarifying. Let's merge (but it needs an approval first from you...)

TheLostLambda (approved):
Whoops, realised I actually need to approve these changes!

PR #350, opened 2021-02-07
Adding a place for 'one-off' scripts

verdammelt:
Thought it might be useful to have a place to put those scripts we may only use once or twice. Perhaps we might see similarities and be able to pull out more general purpose utilities from the.


TheLostLambda (approved):
I think this is a great idea! Any scripts for converting from cl-unit to fiveam could probably also live here.

verdammelt:
I was hoping there would be agreement that this sort of thing is useful. We'll be able to at least copy-n-paste from old solutions to new solutions, or better create a utility library for maintenance scripts over time.

PR #351, opened 2021-02-07
Switch from cl-json to yason in CI system.

verdammelt:
cl-json does something very strange with the "v2" key and it annoyed
me enough I decided to switch to yason.


TheLostLambda (approved):
Haha, the number of JSON libraries for CL absolutely horrifies me, but this looks like a good way to clean things up!

PR #352, opened 2021-02-07
Quieten the build.

verdammelt:
I don't like noisy builds so I squelched some of the noise.

I worry though that by muffling style warnings (these are the compiler notes from SBCL) we might actually be missing something...

Thoughts?

### DIFF
--- a/src/test-exercises/exercises.lisp
+++ b/src/test-exercises/exercises.lisp
@@ -16,7 +16,8 @@
 (defun load-exercise-file (exercise file)
   (let ((*default-pathname-defaults* (aget :directory exercise))
         (*load-verbose* t))
-    (load file)))
+    (handler-bind ((style-warning #'muffle-warning))
+      (load file))))
 
 (defun find-exercise-package (exercise &key (test nil))
   (let ((slug (symbol-name (aget :slug exercise))))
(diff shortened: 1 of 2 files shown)


TheLostLambda (commented):
Hmm, I'm a bit divided on this as well. Do you have a comparison of the CI outputs?

I think that it's nice to have a quiet, clean build, but warnings are legitimately helpful to have. Maybe warnings could have their own, separate CI job? Or maybe it's a flag for this one?

Ultimately though, I'm fine with whichever you decide to prefer (I am curious about the warnings we are squelching though)

verdammelt:
The style warnings we don't want are the ones creaetd when you load the stub solution file - this file generates a lot of unused variable etc. type message.s

The style warnings I think we want are from the test and example/exampler files.

TheLostLambda:
Ah! I didn't quite catch that nuance the first time around! Yeah, I actually think this is a good idea then. I think if there are legitimate style errors in things like example files (I don't know if those are actually tested here?) that would be something we could catch during local development.

I think this is a good idea for the time being!

TheLostLambda (approved):

verdammelt:
My worry about silencing useful warnings (such as from test files or example/exemplar files) worries me - but hopefully we'd see those in development. But I'll also add to my todo list a sort of 'linter' CI that could go through those files and load them out and perhaps fail if there are warnings, at least report them.

Issue #347, opened 2021-02-05
Practice exercises must be updated to v3 format.


verdammelt:
1. They must be updated to use `fiveam` rather than `lisp-unit` and in the new testing style (no more #'xlisp-test feature checking...)
1. They must conform to the [specification](https://github.com/exercism/docs/blob/main/anatomy/tracks/practice-exercises.md)

## Exercises needing to be updated.

Before working on one - "claim" it then when a PR is up add that detail. Check off when PR merged.

- [ ] acronym
- [ ] all-your-base
- [ ] allergies
- [ ] anagram
- [ ] armstrong-numbers
- [ ] atbash-cipher
- [ ] beer-song
- [ ] binary
- [ ] binary-search
- [ ] bob
- [ ] collatz-conjecture
- [ ] crypto-square
- [ ] difference-of-squares
- [ ] etl
- [ ] gigasecond
- [ ] grade-school
- [ ] grains
- [ ] hamming
- [ ] hello-world (PR: #356)
- [ ] isogram
- [ ] leap
- [ ] luhn
- [ ] meetup
- [ ] nucleotide-count
- [ ] pascals-triangle
- [ ] perfect-numbers
- [ ] phone-number
- [ ] prime-factors
- [ ] raindrops
- [ ] rna-transcription
- [ ] robot-name
- [ ] robot-simulator
- [ ] roman-numerals
- [ ] scrabble-score
- [ ] sieve
- [ ] space-age
- [ ] strain
- [ ] sublist
- [ ] triangle
- [ ] trinary
- [ ] twelve-days
- [ ] two-fer
- [ ] word-count

verdammelt:
There are 43 practice exercises to update.

We could put a big todo list here with all the exercises... or we could replace this issue with an issue per exercise.

Thoughts?

verdammelt:
For now i'll put in a list of exercises that need updating.

TheLostLambda:
I think that this one issue is a much cleaner approach! I like this quite a bit! Thank you for putting this together!

verdammelt:
Reviewing issues I see that this was done - so closing.

PR #356, opened 2021-02-12
Practice update/hello world

verdammelt:
This PR updates the hello-world practice exercise to v3.

Along the way I found out that the CI system assumed `exemplar` files and couldn't handle the practice exercise `example` file config key (this had been adjusted for v2 tests - seem to have overlooked doing this for v3 tests.

(ref: #347)

### DIFF
--- a/exercises/practice/hello-world/hello-world-test.lisp
+++ b/exercises/practice/hello-world/hello-world-test.lisp
@@ -1,20 +1,27 @@
-;;; 
-;;; hello-world v1.1.0
-;;; 
-(ql:quickload "lisp-unit")
-#-xlisp-test (load "hello-world")
+;; Ensures that key-comparison.lisp and the testing library are always loaded
+(eval-when (:compile-toplevel :load-toplevel :execute)
+  (load "hello-world")
+  (ql:quickload :fiveam))
 
-(defpackage #:hello-world-test
-  (:use #:common-lisp #:lisp-unit))
-(in-package #:hello-world-test)
+;; Defines the testing package with FiveAM in scope
+;; The `run-tests` function is exported for use by both the user and test-runner
+(defpackage :hello-world-test
+  (:use :cl :fiveam)
+  (:export :run-tests))
 
-(define-test
-  say-hi!
-  (assert-equal
-    "Hello, World!"
-    (hello-world:hello)))
+;; Enter the testing package
+(in-package :hello-world-test)
 
-#-xlisp-test
-(let ((*print-errors* t)
-      (*print-failures* t))
-  (run-tests :all))
+;; Define and enter a new FiveAM test-suite
+(def-suite hello-world-suite)
+(in-suite hello-world-suite)
+
+(test say-hi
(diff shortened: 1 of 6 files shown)


exercises/practice/hello-world/hello-world-test.lisp
```diff
@@ -1,20 +1,27 @@
...
-      (*print-failures* t))
-  (run-tests :all))
+;; Define and enter a new FiveAM test-suite
+(def-suite hello-world-suite)
+(in-suite hello-world-suite)
+
+(test say-hi
+      (is (string= "Hello, World!" (hello-world:hello))))
+
+;; Either provides human-readable results to the user or machine-readable
```
TheLostLambda, line 22: The latest version of the test runner doesn't require this! If you'd like, we can leave this bit of repetitive code out of all of our tests going forward. Let me know if you'd like me to remove it from the template as well!

TheLostLambda (approved):
Looking good! We can simplify the tests a little bit now!

Do you think this process is easyish to automate? Or should be just power through it by hand?

exercises/practice/hello-world/hello-world-test.lisp
verdammelt: Sorry - what bit of repeative code? the run-test function?

verdammelt:
RE: automating - I don't know if it would be easy to automate unless we built a new exercise generator first. And I'm a bit torn if we should get practice exercises ready for v3 vs. spending the time on the generator up-front.

TheLostLambda:
How does the current generator work? How much work do you think it would be to put together something like that that's a bit more future-proof?

verdammelt:
Current generator works for v2 exercises - so would need to be updated to v3 practice exercises - later we could add in v3 concept exercises.

it didn't use template files like your version that you wrote for v3 pre-big-bang. If I was going to update it I'd probably also want to do that - easier to change when it is not part fo the code.

Maybe I'll try to do that today (today is a Holiday here in US).

TheLostLambda:
> Maybe I'll try to do that today (today is a Holiday here in US).

Sounds stellar! Let me know if you need any help from me! I've got a bit of a busy day, but I've done some thinking about features and might start on adding prerequisites!

exercises/practice/hello-world/hello-world-test.lisp
TheLostLambda: Yeah! Sorry for not clarifying! We are just directly calling the tests by parsing the AST and don't need to bother with any sort of run-test function anymore!

verdammelt: I actually like the run-tests... but I can simplify it down to just `explain!` on the suite.

I'm working on a little one-off script to take a v2 exercise and make it v3 - so we can use that for the heavy lifting of changing the exercises at lease.

verdammelt:
Please look to #358 where-in I wrote a hacky script for converting exercises and did it to `robot-simulator`. If we like how it worked out then I suggest closing this PR *and* that PR and using the script to convert exercises in bulk.

PR #361, opened 2021-02-19
Update instructions to not allow 0 area triangle

sideshowcoder:
The test cases for this exercise contain a test case to disallow 0 area triangles

```
(test invalid-2 (is (equal :illogical (triangle:triangle 1 2 1))))
```

It is arguable what is correct, a triangle with `a + b = c` has an area of 0, but is correct in terms of the triangle equation. I choose to update the instructions to only allow `a + b > c`.

### DIFF
--- a/exercises/practice/triangle/.docs/instructions.md
+++ b/exercises/practice/triangle/.docs/instructions.md
@@ -13,8 +13,8 @@ A _scalene_ triangle has all sides of different lengths.
 ## Note
 
 For a shape to be a triangle at all, all sides have to be of length > 0, and
-the sum of the lengths of any two sides must be greater than or equal to the
-length of the third side. See [Triangle Inequality](https://en.wikipedia.org/wiki/Triangle_inequality).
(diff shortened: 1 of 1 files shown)


verdammelt:
Thanks @sideshowcoder for noticing this.

A note to who-ever works on this - update the exercise from problem-specifications canonical data and see if the problem lies with that rather just in this track.

verdammelt:
Problem-specifications discussion that appears relavent: problem-specifications/#202

kotp:
Degenerative triangles are being used in at least several tracks on Exercism.  It does stem from problem-specifications, but I think the mentioned issue is not the only discussion about it.

verdammelt:
After reviewing the current problem specifications and discussion around this exercise I think I will close this PR but add an issue to update this exercise from problem specifications and make it a priority. It is generally easier for me to keep practice exercises (generally) in sync with the specifications so i'd like to just do that.

Sorry it took me so long to get around to this PR and making a decision.

PR #362, opened 2021-02-20
Ensure all exercises have a UUID.

verdammelt:


TheLostLambda (approved):
Looks pretty good an I'm eager to get this fixed, but does beta mean these non-existent exercises will be visible to everyone? Maybe it would be better if they were marked WIP so only we saw them.

I think that's how those tags worked, but I can't find the docs at the immediate moment.

verdammelt:
Good catch about the status - the docs seem to agree that we should set those statuses to WIP and so i did.

TheLostLambda:
Looks great! Thanks for doing this!

iHiD:
A step forward! Nice work, both :)

PR #378, opened 2021-04-05
Cleanup concepts

verdammelt:
This PR does not intend to be "the last word" on concepts for this track. Its intention is to clean up concepts to the point where the build will be green and the track will be in a consistent state with respect to concepts.

Along the way a "quick and dirty" config checker to do checks in addition to configlet was developed. This checker can be improved and expanded as desired.


verdammelt:
Will leave this open for a few days before merging.

PR #420, opened 2021-07-04
Fixing problems with text of the concept and exercise.

verdammelt:
The about.md was only an additional note rather than a recapitualation
with additional note. The blurb was for an earlier idea for the
exercise. The introduction talked about character codes but the
exercise no longer used it.

### DIFF
--- a/concepts/characters/introduction.md
+++ b/concepts/characters/introduction.md
@@ -12,13 +12,6 @@ Some non-graphical characters such as space or new-line have names such as `#\Sp
 The standard only requires an implementation to have 96 characters (upper and lower case Latin alphabetic characters (`A-Za-z`), the digits (`0-9`), space (`#\Space`), newline (#\Newline) and punctuation (e.g. ``!$"'(),_-./:;?+<=>#%&*@[\]{|}`^~``)
 However most implementations will offer more than that, including implementations which provide all Unicode characters.
 
-## Character codes
-
-Each character has a 'code' which is a non-negative integer which is unique to that character in a specific implementation.
-For example:
-
-(Note the mapping of character to integer is dependent upon implementation so your implementation may have the same or different codes as the examples.)
-
 ## Comparing characters.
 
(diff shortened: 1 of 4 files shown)


exercises/concept/character-study/.meta/config.json
```diff
@@ -5,5 +5,5 @@
     "test": ["character-study-test.lisp"],
     "exemplar": [".meta/exemplar.lisp"]
   },
-  "blurb": "Learn the basics of Common Lisp characters by working on cyphers"
+  "blurb": "Learn the basics of Common Lisp characters by helpeing Lewis learn about Earth letters."
```
TheLostLambda, line 8: It seems that all of Exercism is a little inconsistent about periods at the ends of blubs – most seem to not have a period at the end, but I've also found a couple that do.

It's a small style thing I'll leave up to you!

concepts/characters/introduction.md
```diff
@@ -12,13 +12,6 @@ Some non-graphical characters such as space or new-line have names such as `#\Sp
 The standard only requires an implementation to have 96 characters (upper and lower case Latin alphabetic characters (`A-Za-z`), the digits (`0-9`), space (`#\Space`), newline (#\Newline) and punctuation (e.g. ``!$"'(),_-./:;?+<=>#%&*@[\]{|}`^~``)
```
TheLostLambda, line 12: `#\Newline` could use some code tags here :)

TheLostLambda (approved):
Overall LGTM! Some missing code tags if you want to fix that up too :)

exercises/concept/character-study/.meta/config.json
verdammelt: also i can't spell 'helping' it seems!

verdammelt: Appears that at least for this track using a period may be more common so I'll keep the period in this case.

PR #421, opened 2021-07-04
Icon config key was wrong.

verdammelt:
The key is "icon" not "icons".


TheLostLambda (approved):
Nice and easy! This matches the V3 docs now :)

PR #423, opened 2021-07-07
🤖 Simplify `hello-world` stub

ErikSchierboom:
Simplify the `hello-world` exercise's stub. The goal of this is to make things easier for students to get started and to introduce as little syntax as possible.

See https://github.com/exercism/v3-launch/issues/46

### DIFF
--- a/exercises/practice/hello-world/hello-world.lisp
+++ b/exercises/practice/hello-world/hello-world.lisp
@@ -1,6 +1,7 @@
 (defpackage #:hello-world
-  (:use #:cl)
+  (:use #:common-lisp)
   (:export #:hello))
+
 (in-package #:hello-world)
 
-(defun hello () "Put your greeting here.")
+(defun hello () "Goodbye, Mars!")
--- a/exercises/practice/hello-world/hello-world.lisp
+++ b/exercises/practice/hello-world/hello-world.lisp
@@ -1,5 +1,5 @@
 (defpackage #:hello-world
-  (:use #:common-lisp)
+  (:use #:cl)
   (:export #:hello))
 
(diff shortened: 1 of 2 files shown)


exercises/practice/hello-world/hello-world.lisp
```diff
@@ -1,6 +1,7 @@
 (defpackage #:hello-world
-  (:use #:cl)
+  (:use #:common-lisp)
```
verdammelt, line 2: I'm curious about this change. `COMMON-LISP` correct - but so is `CL` (a nickname for `COMMON-LISP` package. I think (or would like to think) that we've standardized on using `CL` on the track.

TheLostLambda: My votes goes for `CL` as well! Otherwise this looks good to me!

ErikSchierboom: I just use the example file as the basis for this PR, so I'll update the example file too.

verdammelt: Whoops! I guess I need to grep the track for `(:use #:common-lisp)`... thanks for bringing it to our attention.

verdammelt (approved):

ErikSchierboom:
I'll wait for @TheLostLambda to approve too.

TheLostLambda (approved):
Looks good to me!

TheLostLambda:
I think I'll merge this now so I can change the rest of the track to use `CL` too! :slightly_smiling_face: 

Thanks to both @ErikSchierboom and @verdammelt !

PR #431, opened 2021-07-15
Move to a simpler test function in existing exercises

### LINKED ISSUE #417: Simplify the V3 Test Files
The V3 `run-tests` function no longer needs to be used by the test-runner, only the learner. We could simplify all of the V3 tests to have the same `run-tests` function as V2 – making things consistent and simpler.

Here is the proposed replacement:

```lisp
(defun run-tests (&optional (test-or-suite 'robot-simulator-suite))
  "Provides human readable results of test run. Default to entire suite."
  (run! test-or-suite))
```

TheLostLambda:
Closes #417

### DIFF
--- a/exercises/concept/character-study/character-study-test.lisp
+++ b/exercises/concept/character-study/character-study-test.lisp
@@ -39,9 +39,6 @@
   (is (eq :newline (type-of-char #\Newline)))
   (is (eq :unknown (type-of-char #\!))))
 
-;; Either provides human-readable results to the user or machine-readable
-;; results to the test runner. The default upon calling `(run-tests)` is to
-;; explain the results in a human-readable way
-(defun run-tests (&optional (explain t))
-  (let ((tests (run 'character-study-suite))) ; Run the tests once
-    (if explain (explain! tests) tests))) ; Optionally explain the results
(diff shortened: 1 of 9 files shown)


TheLostLambda:
@verdammelt Did we lose the exercise generator somewhere along the way? The one that took the template files and generated the scaffolding for a new concept exercise?

verdammelt:
Not "lost" but I doubt it works correctly anymore. I think I opened a issue to make a new one or get the existing working again.

TheLostLambda:
Ah, that's fair. I just couldn't find it anywhere in the repository! That makes more sense :)

verdammelt:
Looks like I removed the generator in commit 6833bc3.

verdammelt (approved):
This looks good to me.

TheLostLambda:
> Looks like I removed the generator in commit [6833bc3](https://github.com/exercism/common-lisp/commit/6833bc31baa22b9451e188ae6eacd1289f1a2be5).

Huh, I don't even think this was the one I was thinking about! I think I had in mind the little template thing I wrote just for v3 concept exercises! I suppose we have a few generators kicking around now :sweat_smile: 

Probably best to continue this hunt elsewhere though as I think I'll merge this :)

Issue #355, opened 2021-02-12
Add shared files to track


verdammelt:
According to the [spec](https://github.com/exercism/docs/blob/main/building/tracks/shared-files.md) this track should have some shared files such as:

- [ ] help.md (#430)
- [ ] tests.md (#430)
- [ ] debug.md (optional) (#434)

The first two are for use on the command line and the last is for use with the in-browser editor.

verdammelt:
completed...

PR #434, opened 2021-07-25
Add instructions on using print to debug test runs.

verdammelt:
Just a quick document explaining how to use `print` to do some debugging for test runs in the browser.

Relates to #355.

### DIFF
--- a/exercises/shared/.docs/debug.md
+++ b/exercises/shared/.docs/debug.md
@@ -1,0 +1,8 @@
+# Debug
+
+To help with debugging, you can use [`print` or similar functions](http://www.lispworks.com/documentation/HyperSpec/Body/f_wr_pr.htm) to write to standard output which will be visible in the test output.
+
+For example `(print "hello there")` in your code will cause it to be visible in the test runs.
+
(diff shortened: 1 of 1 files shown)


verdammelt:
@TheLostLambda can you confirm that in fact `print` output will be visible in the test output when run via the `test-runner`?

TheLostLambda:
> @TheLostLambda can you confirm that in fact `print` output will be visible in the test output when run via the `test-runner`?

I'm pretty sure, but I'll double check now!

TheLostLambda:
I can confirm that code like this:
```lisp
;; Evaluates to T if THING is an atom, NIL otherwise
(defun is-an-atom-p (thing) (print thing))

;; Evaluates to T if THING is a cons, NIL otherwise
(defun is-a-cons-p (thing) (format t "Is this a cons? ~A~%" thing))
```

Generates outputs like this:
image
and
image

TheLostLambda (approved):
It might be worth saying that any printing **needs** to be done in a tested function. If the print statements are just slapped in the top-level file somewhere, they won't show up anywhere.

Dynamic scoping, however, means that it's fine to put print messages in helper functions that are ultimately called by the tested functions.

verdammelt:
@TheLostLambda I'll tweak the wording as you suggest.

PR #435, opened 2021-07-28
Add icons for key features

ee7:
The icons for key features are now available, and `configlet lint` now requires that each key feature has a valid `icon` value.

This PR sets an initial icon for each key feature - feel free to change them however you want. You can see the list of available icons [here](https://github.com/exercism/docs/issues/189). Note that an icon can be used regardless of its name.

With the initial state of this PR, the key features look something like the below.

---

interactive Interactive
Common Lisp has a full featured REPL, development can happen as a 'conversation' with the system.

documentation Inspectible
Common Lisp is fully inspectable with built-in functions for exploring every part of the language.

homoiconic Homoiconicity
Code has the same structure as data, allowing for powerful macros: code that writes code.

extensible Extensible
Generic functions, reader-macros, and flexible namespacing allow for the extension of the language

stable Standardized
The language is very stable, with multiple implementations to match your needs – JVM, embedded, etc.

multi-paradigm Multiparadigm
Whether you prefer a functional style or something more object-oriented, Lisp adapts to your needs.

### DIFF
--- a/config.json
+++ b/config.json
@@ -829,32 +829,32 @@
   ],
   "key_features": [
     {
-      "icon": "feature-interactive",
+      "icon": "interactive",
       "title": "Interactive",
       "content": "Common Lisp has a full featured REPL, development can happen as a 'conversation' with the system."
     },
     {
-      "icon": "feature-inspectable",
+      "icon": "documentation",
       "title": "Inspectible",
       "content": "Common Lisp is fully inspectable with built-in functions for exploring every part of the language."
     },
     {
-      "icon": "feature-homoiconicity",
+      "icon": "homoiconic",
       "title": "Homoiconicity",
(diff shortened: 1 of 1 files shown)


verdammelt:
@ee7 I believe this change is already on the main branch. The left-side of the diffs in this PR is showing an older version of the config.json. Please update this PR with latest code from the main branch so we can see if there are any actual changes. Thanks.

ee7:
As you say, the `common-lisp` track already added the icons. There's no important diff - I'm closing this PR. Sorry for the noise.

This happened because I started working on the icons a bit earlier, and I had a comically bad/non-existent internet connection for a while - I couldn't pull the new changes before pushing.

I picked the icons for 20-ish tracks, and then ran a script to create the PRs. It took more than an hour for it to finish :)

verdammelt:
Excellent thanks for the PR none-the-less.

PR #437, opened 2021-08-20
Standardise on `:` over `#:`

### LINKED ISSUE #427: Standardise on either interned `:cl` or uninterned `#:cl` symbols in pacakge definitions
I personally find `:cl` a bit less verbose / noisy, but `#:cl` might be better practice to avoid polluting the package namespace?

We should do a bit of reading up on best-practices and then make things consistent!

TheLostLambda:
Closes #427 (if I didn't mess up something important...)

### DIFF
--- a/docs/template-production.lisp
+++ b/docs/template-production.lisp
@@ -1,9 +1,9 @@
-(in-package #:cl-user)
-(defpackage #:exercise
-  (:use #:cl)
-  (:shadow #:list)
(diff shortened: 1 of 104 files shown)


verdammelt (approved):
Change looks good - thanks for doing this.

I'll make similar changes in the representer & analyzer so we have a consistent coding style across the repos.

PR #470, opened 2021-09-01
Typographical error and trailing newlines

kotp:


verdammelt (approved):
Good catches. Thanks.

PR #475, opened 2021-09-04
Minor fixes in Leslie's list's instructions and template solution

alvaro121:
Add missing paramater name to function template and fix using wrong function in the instructions.

Also, in the _Bigger lists out of smaller lists_ section, it says:
> Leslie realized they accidentally made two shopping lists not one! Write a function called list-append which adds all the items from the second list provided to the end of the first list.

I assume the intended solution would be to use `append` and not `nconc`, but that would not _add the items to the list_ but instead return a new list.

### DIFF
--- a/exercises/concept/leslies-lists/.docs/instructions.md
+++ b/exercises/concept/leslies-lists/.docs/instructions.md
@@ -64,5 +64,5 @@ Leslie realized they accidentally made two shopping lists not one! Write a funct
 Leslie is starting to get worried that this shopping trip is going to take quite a while. Just how many things are on this list? Write a function `just-how-long` to tell them just how long their list is.
 
 ```lisp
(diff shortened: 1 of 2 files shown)


verdammelt:
Thanks for catching these.

As for the append vs. nconc issue - you bring up a good point. But I want to steer the student to append rather than nconc at least at the beginning. Perhaps we can reword the story to make it clear that we are not mutating a list but returning a new one. We could also put a mention of nconc in the about.md which is shown after an exercise is completed and can give additional information about a concept.

I'll open an issue about that to improve the exercsise and feel free to create another PR for it if I don't get to it first.

verdammelt (approved):

verdammelt:
Thanks again for your contribution to the track!

PR #480, opened 2021-09-04
Correct evaluation result

goderich:


verdammelt:
Thanks, but in reviewing this I see that the text has a bigger problem.

The text above this example (in both about.md and introduction.md) says that `remove-if` will return a list for which the predicate is true - but in fact it should be 'for which the predicate is false'.

Also this specific example should be reworked so it shows that an empty list is returned. So it is the predicate which is wrong.

verdammelt:
@goderich what do you think of the change i just pushed on this branch?

goderich:
Yeah, no problem, I was just taking a quick look at v3 and wanted to bring the mistake to your attention, but didn't have time to evaborate, sorry.

I suggest editing the description though. I personally find it a bit difficult to understand, even though I know what the function does. Right now you have:

> `remove-if` is a function that takes a function and a sequence and evaluates to a sequence of those elements for which the predicate evaluated to a false value.

How about changing this to something like:

>`remove-if` is a function that takes a predicate function and a sequence as arguments, and returns a sequence of only those elements for which the predicate evaluated to a false value.

I also think explicitly stating that the function does not modify the original list would be helpful, especially because in CL it's not always obvious when that happens.

>In other words, it "removes" all elements of the sequence for which the predicate evaluates to true. However, note that this function does not modify the original sequence! Instead, it returns a new sequence with the appropriate elements removed.

verdammelt:
Thanks for your help in improving this concept @goderich

PR #481, opened 2021-09-04
Add missing "no" at conditional concept

alvaro121:
The meaning of the sentence was inverted. The sentence was repeated in two places, I changed both.

Should it be mentioned that if more than one body expression is provided the last one is the overall result (and the others are executed for side-effects)?

I also noticed that the corresponding about.md does not mention the possibility of providing more than one body expression, is this important?

### DIFF
--- a/concepts/conditionals/about.md
+++ b/concepts/conditionals/about.md
@@ -16,6 +16,8 @@ As an example:
 
 The conditionals `when` and `unless` evaluate some code only when the provided test is true or false respectively – evaluating to `nil` otherwise.
 
+The section after the test expression may be more than one expression.
+
 ```lisp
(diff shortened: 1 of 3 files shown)


verdammelt:
Thanks for catching this. No matter how many times i've read through these docs I just never see things like this.

The introduction to the exercise is created via a generator from the concept docs so once we get teh concept introduction as we want it then we can regenerate the exercise intrroduction.

As for the about.md - yes I think it should also mention the multi-expression ability. The about.md is shown to the student after completing the exercise and is *the* information shown if the student wants to review the concept.

I'll make those changes - add it into this PR, approve and merge it.

verdammelt (approved):

PR #483, opened 2021-09-05
Typo fix at mapping concept and nconc specification

alvaro121:
There was a typo in the specified functions. Also I clarified that `nconc` modifies all the lists but the last one; this is not specified verbatim at the [CLHS](http://www.lispworks.com/documentation/HyperSpec/Body/f_nconc.htm) but I think is right, is it?

Feel free to make any edits.


verdammelt (approved):
Thanks. I had not read `nconc` carefully when writing this file. I was going from memory.

Issue #476, opened 2021-09-04
List concept exercise: nconc vs. append


verdammelt:
Quoting @alvaro121 in #475:

> Also, in the Bigger lists out of smaller lists section, it says:
> 
> Leslie realized they accidentally made two shopping lists not one! Write a function called list-append which adds all the items from the second list provided to the end of the first list.
> I assume the intended solution would be to use append and not nconc, but that would not add the items to the list but instead return a new list.

I suggest we correct the story text to not imply mutating the list and add a mention of nconc to the about.md for the list concept.

alvaro121:
Is there a _mutation_ concept or similar? I could only find #318 that was closed. The would-be concept seems to be referenced at the [hash table concept](https://github.com/exercism/common-lisp/blob/c5eee935a3c8490257b5ca4a8e5c0c3375001d78/concepts/hash-tables/introduction.md):
> The use of `setf` will be explained in another concept.

verdammelt:
This is currently no mutation concept. The `setf` concept is going to be an interesting one - perhaps several conepts, as there is first simple variable assignment, then the topic of "places" and then discussion of the 3(or more) ways to define a `setf` function.

PR #488, opened 2021-09-07
Remove mutation implication at leslies-list instructions

alvaro121:
Removed the implication of mutation in the instructions as mentioned in #476.

I avoided phrasing the returned value as a _new list_ due to [`append`](http://www.lispworks.com/documentation/HyperSpec/Body/f_append.htm) returning its last argument directly "if there are no preceding non-empty lists".

@verdammelt, I couldn't find the `about.md` file you referred at #475, did you meant `concepts/lists/about.md`?


verdammelt (approved):
Looks good. Thanks for the editing.

verdammelt:
RE: about.md - yes the concept about.md. We can do something about that (if we choose to) as a separate piece of work (#476)

PR #495, opened 2021-09-07
Update robot-simulator.lisp

mtreis86:
Default functions didn't exist

### DIFF
--- a/exercises/practice/robot-simulator/robot-simulator.lisp
+++ b/exercises/practice/robot-simulator/robot-simulator.lisp
@@ -4,3 +4,11 @@
            :robot :robot-position :robot-bearing :make-robot))
 
 (in-package :robot-simulator)
+
+;; A robot has a position on a grid made up of a cons of positive numbers,
+;; and a bearing of +north+ +south+ +east+ or +west+
+(defstruct robot
+  position bearing)
+
+(defun execute-sequence (robot sequence)
+  "Moves and rotates the robot according to the sequence")


mtreis86:
I don't really understand why this is failing, is robot-bearing defined elsewhere?

verdammelt:
Part of the CI is to run the exercise with the solutions we provide in the example or exemplar files. Looks like this is where it is failing. The message is:

> 2021-09-07T20:21:59.3020549Z ;;; >>>FAILURE: Cannot CHANGE-CLASS objects into CLASS metaobjects.
> 2021-09-07T20:21:59.3020990Z See also:
> 2021-09-07T20:21:59.3021387Z   AMOP, Initialization of Class Metaobjects
> 2021-09-07T20:21:59.3022015Z ;; PRACTICE ROBOT-SIMULATOR (PASS 0 FAIL 0 ERRORS 1)

mtreis86:
I am also noticing the symbols, +north+ +south+ etc don't make sense to me, in the test they're all unbound and also not defined in the problem. Do you have a working solution to this puzzle we can use for the template?

mtreis86:
Oh nevermind I see it in the .meta folder, I gotcha

verdammelt:
i'm not sure if this exercise is a good one for stubs. You are proposing to use a structure. The example file suggests a class. Neither is more correct than the other. One could implement this with a set of functions and a simple list.

PR #502, opened 2021-09-09
Switching CI to use clfoundation images.

### LINKED ISSUE #478: Use cl-foundation images for CI actions?
Currently CI actions such as test-exercises and config-checker are run in a generic image and we have to install SBCL etc. can we use [clfoundation images][https://hub.docker.com/r/clfoundation/sbcl] instead?

verdammelt:
Still need to do some steps to set up the image - but now more of it is done by someone else.

Fixes #478

### DIFF
--- a/.github/workflows/config-checker.yml
+++ b/.github/workflows/config-checker.yml
@@ -6,27 +6,24 @@ jobs:
   config-check:
     runs-on: ubuntu-latest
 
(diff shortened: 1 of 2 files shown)


verdammelt:
Might be nice to pull out an action to do test-system on a given system - but that could be done some other time.

PR #505, opened 2021-09-09
Change instructions to be consistent with tests

stutonk:
The tests for this exercise specify a `:small` keyword, among others, with respect to the "size" of a character. This was incorrectly listed as `:little` in section 2 of the instructions.


verdammelt (approved):
Good catch.

PR #511, opened 2021-09-12
Fix typos

wsgac:
Hey there. I found some inconsistencies between function names in the docs vs. those used in the source files.


verdammelt (approved):
Thanks for catching these.

PR #516, opened 2021-09-16
functions: Fixed the example function name

ZapAnton:
Replaces the `+1` with the `1+` function call


verdammelt:
Good catch - thanks.

PR #532, opened 2021-10-31
fix typo

fu7mu4:
event number -> even number


verdammelt (approved):
Thanks for catching this!

PR #535, opened 2021-11-10
Fixes typo 'later' to 'latter'

ketigid:
Submission to fix a typo noticed while doing the exercise.


verdammelt:
This looks good but it is not fixing it in the correct place. This file is generated rom the `introduction.md.tpl` file which merges the date-time and multiple-value concept document files. So the fix needs to actually happen in the date-time document and then this file re-generated.

If you are up to date - then go for it. Otherwise I'll take care of it myself later...

verdammelt (changes requested):
Looks good - but fixes the problem in the generated file rather than the source file.

verdammelt (approved):

PR #536, opened 2021-11-19
Specify that Lenny's favorite food is not null

### LINKED ISSUE #469: Improve Sorting Socks and Sexpr exercise
I just stumbled upon the fact that one does not need to actually write anything for the first function. It passes all of its tests wihtout changing the stub: `(defun lennys-favorite-food ())`

djuber:
The original test for `symbolp` and not `keywordp` passed with no
action, because the empty function body returned nil, and `(symbolp
nil)` is true.

Require a non-nil symbol response for `(lennys-favorite-food)`.

Fixes #469


TheLostLambda (approved):
An excellent, simple solution! Thanks for the contribution!

PR #537, opened 2021-11-23
Create format-basics concept & concept exercise.

### LINKED ISSUE #447: Add 'format' concept & exercise
The text-formatting stub concept was removed before launch in #443.

This concept should teach the basics of how to use 'format'. Would be good to allow it to be a relatively early concept so it should not delve into streams as those are likely to be a much later concept.

Perhaps the exercise would just use `nil` stream for creating strings.

There will likely be a whole tree of 'format' concepts as it is a complex topic.

verdammelt:
See #447 for details about proposed design.

Fixes #447

### DIFF
--- a/concepts/format-basics/.meta/config.json
+++ b/concepts/format-basics/.meta/config.json
@@ -1,0 +1,4 @@
+{
+  "blurb": "Basic information about formatting output.",
+  "authors": ["verdammelt"]
+}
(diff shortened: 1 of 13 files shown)


concepts/format-basics/.meta/config.json
```diff
@@ -0,0 +1,4 @@
+{
+  "blurb": "Basic information about formatting output.",
+  "authors": ["verdammelt"]
```
verdammelt, line 3: testing how this github-review package works...

PR #541, opened 2021-12-08
Fixes typo in introduction.md

ketigid:
- Fixes `and` to `a`


verdammelt:
Thanks for catching this.

The `introduction.md` file is generated however from the `introduction.md.tpl` file which pulls in the text from the introduction files of the concepts.

So this problem is actually in the text of the concept.

If you could please fix the problem in documentation for the `format-basics` concept and then run `configlet generate` to get the exercise's documentation regenerated. 

Thanks.

ketigid:
Thanks for explaining. Hope I've got it right.

verdammelt (approved):
Looks good Thanks.!

PR #545, opened 2022-01-11
Fixed typo

PaulT89:
Minimal change proposed.  However, I was also toying with the idea of rewriting the sentence: 

 **'In Common Lisp a function can have some arguments that are optional.'**

as something like:

   **'In Common Lisp, it is possible for functions to take optional arguments.'**

or maybe:

   **'Common Lisp allows the user to add optional arguments to their functions.'**

### DIFF
--- a/concepts/optional-parameters/introduction.md
+++ b/concepts/optional-parameters/introduction.md
@@ -1,6 +1,6 @@
 # Introduction
 
-In Common Lisp a function can have some arguments are are optional.
+In Common Lisp a function can have some arguments that are optional.
 These are designated in the lambda list by `&optional` lambda list keyword.
 A parameter will be bound to the value `nil` if it is not specified.
 If there are several optional parameters they are bound in order.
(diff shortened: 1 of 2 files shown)


verdammelt:
Other than the typo (are/that), what is it about that sentence that you feel is unclear or incorrect?

PaulT89:
It's not that the original sentence is _unclear_ so much as (to me at least) there's something about it that doesn't fit as neatly into its surroundings as I think it should.  I think it might be the fact that it's written in a kind of passive way - as though functions with optional parameters are simply things to be encountered by the user, rather than tools that the user can implement and use themselves (which is what the concept is trying to teach, after all).

However, I understand that this is probably a style over substance issue, which is why I only put the rewrites in the comments section.

verdammelt:
Passive speech writing is probably something I do too much of... we'll definitely merge (later... at work now...) the change you made and I'll review your suggestions again.

Thanks!

verdammelt:
@PaulT89 I used this sort of phrasing in several places about arguments: keyword arguments, optional arguments, lambda lists.

So I think for now I am going to merge this PR. Feel free to open an issue or PR to edit the text more here and in other places.

Thanks again.

PR #564, opened 2022-01-24
Include sync output in created issue

verdammelt:
When the configlet-sync workflow runs and detects out of sync exercises we want the exercise to include the output as it contains the information needed to fix the issue.

This PR includes the needed change to the workflow and template to do this.

To communicate the output between the syncing step and issue creation step we write the output directly to an environment variable in `$GITHUB_ENV`. We do this for two reasons: 1) it helps work around a problem with multi-line outputs (ref: https://trstringer.com/github-actions-multiline-strings/) and 2) we need an environment for the injection of values into the template.

### DIFF
--- a/.github/configlet-sync-issue.md
+++ b/.github/configlet-sync-issue.md
@@ -4,4 +4,10 @@ title: Out of sync practice exercises
 
 # Out of sync practice exercises
 
+This is the new text...
+
 `configlet sync --tests --docs --metadata --filepaths` found some out of sync practice exercises.
+
+```
+{{ env.SYNC_OUTPUT }}
+```
--- a/.github/configlet-sync-issue.md
+++ b/.github/configlet-sync-issue.md
@@ -2,12 +2,11 @@
 title: Out of sync practice exercises
 ---
 
-# Out of sync practice exercises
+Out of sync practice exercises have been detected:
 
-This is the new text...
+Command: `configlet sync --tests --docs --metadata --filepaths`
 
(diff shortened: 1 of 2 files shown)


ErikSchierboom (commented):
I like the idea and the simplicity of the implementation.

My only comment would be: what to do when an existing issue already exists? Ideally, the existing issue either get updated or a comment is posted I think. You probably don't want to have multiple of the same issues.

verdammelt:
@ErikSchierboom that's a feature of the create-an-issue action I'm using. The [`update_existing`](https://github.com/exercism/common-lisp/blob/387203d96ec4baa88640d5d35de0a970ea01b163/.github/workflows/configlet-sync.yml#L35) tells the action to look for existing open issues with the same name and then update it. I saw this happening during my testing.

There is also a further setting which would tell the action to look for even closed issues - so there would be only a single one of these issues in a track; if it was closed it would be reopened. I did not experiment with that setting.

ErikSchierboom:
That's great! The closing/reopening option would be nice too, if you find the time to experiment it would be much appreciated.

verdammelt:
In testing the re-opening i stumbled upon a bug in the action i'm using - there is an open issue on it from November with a comments that points at a fork that is 'maintained' hmmm will investigate perhaps the action i have chosen is "dead"?

verdammelt:
And the 'maintained' fork also doesn't work.

Well this PR works well enough for me - so i will keep it for now... but it does not resurrect closed issuess which would be cool. But it won't create duplicate issues at least.

PR #591, opened 2022-04-22
Adding practice exercise book-store

PaulT89:
## Summary

Adding a practice exercise.  Made my own generator (sort of) to create tests and description from the problem-specifications. Difficulty level on this is 6 - seems to hover at medium-hard difficulty on other tracks, so this should be fine.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/book-store/.meta/example.lisp
+++ b/exercises/practice/book-store/.meta/example.lisp
@@ -12,19 +12,19 @@
     (apply '+ (mapcar #'calculate-group-price groupings)))
    0))
 
-;;; Function to recursively find the largest groups possible from the basket.
-;;; Returns the length of each group of books.
 (defun create-groups (basket)
+ "Function to recursively find the largest groups possible from the basket.
+Returns the length of each group of books."
   (unless (zerop (list-length basket))
     (let ((group (remove-duplicates basket)))
       (loop for book in group
         do (setf basket (remove book basket :count 1)))
       (cons (list-length group) (create-groups basket)))))
 
-;;; Function to replace every pair of lengths 3 & 5 with pairs of length 4 & 4.
-;;; This is required to correctly calculate the price, as pairs of 4 & 4 are
-;;; less expensive than pairs of 5 & 3.
 (defun improve-groupings (groupings)
+  "Function to replace every pair of lengths 3 & 5 with pairs of length 4 & 4.
+This is required to correctly calculate the price, as pairs of 4 & 4 are less
+expensive than pairs of 5 & 3."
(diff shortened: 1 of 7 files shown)


verdammelt:
Thank you very much for submitting this. I will take a better look soon but wanted to jump in to say that  if you want add a PR to add in this generator of yours... I'd say it should live under `src` and be usable via the REPL

PaulT89:
It's basically a python script, but it's not very user-friendly - there's a bunch of hardcoding, and it doesn't create all the requisite files, just the test file and the instructions/description.  It's got a long way to go before it's even user-friendly for me, let alone for anyone else.

verdammelt:
"before it's even user-friendly for me"

haha I've got scripts like that!

exercises/practice/book-store/.meta/example.lisp
```diff
@@ -12,13 +12,18 @@
     (apply '+ (mapcar #'calculate-group-price groupings)))
    0))
 
+;;; Function to recursively find the largest groups possible from the basket.
```
verdammelt, line 15: Instead of comments perhaps doc strings would be better?

PaulT89: Fixed (I think...)

verdammelt (approved):

verdammelt:
@PaulT89 This looks good to me, if you are unable to merge let me know and I'll press the button.

One thing I noticed is that the problem-specification contains a comment about problem that implementers should likely not use floating point numbers - so should the example implmentation use rationals instead perhaps? Would that work? Would it be better?

PaulT89:
I don't have write access, so you'll have to merge it.  

As for the prob-spec comment regarding floating-point numbers, I have to be honest and say that I never saw it (oops!).  I did wonder about whether or not I should allow answers as floats, and I figured that it didn't matter.  However, since you brought this to my attention, I'll go back and add a `coerce` into the example, and change the `equalp`s in the test to something stricter.

PaulT89:
Okay, done.  Also added you as a contributor.

verdammelt:
You not being able to merge gave me the opportunity to add labels so you'll get a bunch of 'rep' for this work. Thanks again!

verdammelt:
Looks like we should remember to check if the exercise needs any syncing before committing - I was updating other exercises an configlet told me that this new exercise needed syncing!

https://github.com/exercism/common-lisp/commit/7466f3b2a282c4283d530bd6ee7e66bff24179fe#diff-4903e3329aa2ed65fdfa0b00ed3c37de97922bbbd51a74a46d1e657d80a2e5b1

Minor copy issue only.

PR #597, opened 2022-05-04
Adding Practice Exercise Generator

PaulT89:
## Summary

This adds an exercise generator (as a Python script).  It also updates the common-lisp/README.md file with instructions on how to use it.

## Checklist
- [x] CI is green

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,29 +62,30 @@ Refer to the document on the [anatomy of a practice exercise][practice-exercise]
 Many practice exercises are part of a canonical set of exercises shared across tracks (information on this can be found in the [problem specifications repository][problem-specs].
 There is a generator in the ./bin folder that you can use to generate all of the requisite files from the problem-specifications.
 (Note, you _will_ need to have cloned the [problem specifications repository][problem-specs] for the generator to work.)
-You can run the executable directly by double-clicking on it and following the prompts, or you can run it from the command line.
+The generator is written in Python, and you will therefore need to have Python 3.8 or later installed.
+You can run the script directly and follow the prompts, or you can run it from the command line.
 If you wish to run the generator from the command line, first navigate to your common-lisp repository.
 From here, there are two ways to run the generator, the first way being to enter the following:
 
 ```sh
-./bin/lisp_exercise_generator
+python ./bin/lisp_exercise_generator.py
 ```
 
 and from there, follow the prompts.
 The second way is to type in:
 
 ```sh
-./bin/lisp_exercise_generator [-f] [path-to-prob-spec exercise author]
+python ./bin/lisp_exercise_generator.py [-f] [path exercise author]
 ```
 
 where:
-- path-to-prob-spec is the relative or absolute path to your problem-specifications repository
+- path is the relative or absolute path to your problem-specifications repository
 - exercise is the name of the exercise to be generated
 - author is your Github handle
-- -f is a flag to force overwrite an existing exercise
+- -f is a flag to force overwrite an already existing exercise
 
 Any one of these methods will generate and fill in all the necessary files, with the exception of the .meta/example.lisp file, which you will need to complete yourself.
-**The common-lisp/config.json file will remain unaltered** - you will have to manually alter this file.
+**The common-lisp/config.json file will remain unaltered - you will have to manually alter this file.**
 
 A Common Lisp replacement for this generator will be coming "soon".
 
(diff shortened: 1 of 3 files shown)


PaulT89:
Huh, that's weird - some of my previous commits were added.  Hopefully it doesn't matter - the only 3 files changed are the two Python files and the changes to the README file.

verdammelt:
Likely your `main` branch that you are requesting to be pulled here is out of date to the `main` branch here in the remote repository.  If you update your branch with something like a pull with rebase then you can force push your branch and this PR will be corrected to only contain the correct changes.

Before we go to merge I'd like to make sure it is cleaned up.

README.md
```diff
@@ -56,15 +56,30 @@ The work needed for a concept exercise can be large, feel free to create an [iss
...
+
+#### Practice Exercise Generation
+
 Many practice exercises are part of a canonical set of exercises shared across tracks (information on this can be found in the [problem specifications repository][problem-specs].
-Before creating a new practice exercise please see if there is already a canonical problem defined there, if there is, this track will want to adhere to it rather than implementing something different.
+If you have Python installed, then there is a generator in the ./bin folder that you can use to generate all of the requisite files from the problem-specifications.
+To run it, navigate to your common-lisp repository and enter the following:
+
+```sh
+python ./bin/lisp_exercise_generator.py
```
verdammelt, line 67: Does it matter if it is python 2 or python 3?

verdammelt: And perhaps we can make the script executable and add a `#!` to run python? so then you'd just `./bin/lisp_exercise_generator`...

README.md
```diff
@@ -56,15 +56,30 @@ The work needed for a concept exercise can be large, feel free to create an [iss
...
+```
+
+From there, you will be prompted to provide three further pieces of information.
+(Note, you _will_ need to have cloned the [problem specifications repository][problem-specs].)
```
verdammelt, line 71: We'll want to be clear *where* to clone that repo to...

README.md
```diff
@@ -56,15 +56,30 @@ The work needed for a concept exercise can be large, feel free to create an [iss
...
 Many practice exercises are part of a canonical set of exercises shared across tracks (information on this can be found in the [problem specifications repository][problem-specs].
-Before creating a new practice exercise please see if there is already a canonical problem defined there, if there is, this track will want to adhere to it rather than implementing something different.
+If you have Python installed, then there is a generator in the ./bin folder that you can use to generate all of the requisite files from the problem-specifications.
+To run it, navigate to your common-lisp repository and enter the following:
+
+```sh
+python ./bin/lisp_exercise_generator.py
+```
+
+From there, you will be prompted to provide three further pieces of information.
```
verdammelt, line 70: Is it a big ask to have this data be something that can be passed in as command line arguments?

bin/lisp_exercise_generator.py
```diff
@@ -0,0 +1,340 @@
+import json, toml, os
+from custom_json_encoder import CustomJSONEncoder
+
+TARGET = os.path.abspath(".\\exercises\\practice")
```
verdammelt, line 4: Might be good to use unix style path separators rather than windows ones.

verdammelt (commented):
Just some initial comments. I'll give it a better look and a test hopefully soon.

README.md
PaulT89: Nope.  That's the first of the three prompts - the relative or absolute filepath to the problem-specifications repo.

bin/lisp_exercise_generator.py
PaulT89: Okay

README.md
PaulT89: I used assignment expressions, so it'll have to be Python 3.8+.  However, I figured out how to create an executable (using pyinstaller) so I'll add that and update the README

README.md
PaulT89: I guess I could do that.  The reason why I didn't was simply because I wanted to make it as foolproof as possible, though I suppose I could add similar checks for command line input.

verdammelt: Perhaps we leave it as is for now... I shouldn't ask for changes like this until I've given it a go myself :)

PaulT89: Too late, already added it.

verdammelt:
I have given this a try and it works well in python3. I first had to install `toml` though, which prompted me to relearn how to use a virtual env so I would not install dependencies globally.

The output has some small issues but I would not expect a generator to work well enough that some review is would not be needed. It seemed to do its job well enough that it might be useful.

The `.EXE` however is not useful and I suggest removing it. It is built for a particular OS and chipset so will not work on other machines.

verdammelt (approved):

verdammelt:
Thank you very much @PaulT89 for writing this generator.

PR #600, opened 2022-05-07
Create exercise run-length-encoding

PaulT89:
Adding exercise Run Length Encoding (generated from problem specifications).

Example should be fairly easy to understand as is (though docstring comments can be added if need be).  Difficulty set to 4, though it could be a 3 - it seems to hover around here for other languages.

## Checklist
- [x] CI is green


verdammelt:
Something I may have missed in previous new practice exercises - but here your example.lisp file does not load cleanly - because of the ordering of definitions SBCL produces warnings:

```
; file: /Users/mark/SRC/exercism/common-lisp/exercises/practice/run-length-encoding/.meta/example.lisp
; in: DEFUN ENCODE
;     (RUN-LENGTH-ENCODING::ENCODE-CHUNK RUN-LENGTH-ENCODING::LEN
;      RUN-LENGTH-ENCODING::LETTER)
; 
; caught STYLE-WARNING:
;   undefined function: RUN-LENGTH-ENCODING::ENCODE-CHUNK

;     (RUN-LENGTH-ENCODING::TAKE-WHILE
;      (LAMBDA (RUN-LENGTH-ENCODING::C)
;        (CHAR= RUN-LENGTH-ENCODING::C RUN-LENGTH-ENCODING::LETTER))
;      RUN-LENGTH-ENCODING::PLAIN)
; 
; caught STYLE-WARNING:
;   undefined function: RUN-LENGTH-ENCODING::TAKE-WHILE
; 
; compilation unit finished
;   Undefined functions:
;     ENCODE-CHUNK TAKE-WHILE
;   caught 2 STYLE-WARNING conditions

; file: /Users/mark/SRC/exercism/common-lisp/exercises/practice/run-length-encoding/.meta/example.lisp
; in: DEFUN DECODE
;     (RUN-LENGTH-ENCODING::DECODE-CHUNK RUN-LENGTH-ENCODING::NUM-STRING
;      RUN-LENGTH-ENCODING::LEN RUN-LENGTH-ENCODING::LETTER)
; 
; caught STYLE-WARNING:
;   undefined function: RUN-LENGTH-ENCODING::DECODE-CHUNK

;     (RUN-LENGTH-ENCODING::TAKE-WHILE #'DIGIT-CHAR-P
;      RUN-LENGTH-ENCODING::COMPRESSED)
; 
; caught STYLE-WARNING:
;   undefined function: RUN-LENGTH-ENCODING::TAKE-WHILE
; 
; compilation unit finished
;   Undefined functions:
;     DECODE-CHUNK TAKE-WHILE
;   caught 2 STYLE-WARNING conditions
```

exercises/practice/run-length-encoding/.meta/example.lisp
```diff
@@ -0,0 +1,37 @@
...
+  (if (zerop (length plain))
+    ""
+    (let* ((letter (char plain 0))
+           (len (length (take-while (lambda (c) (char= c letter)) plain))))
+      (concatenate 'string (encode-chunk len letter) (encode (subseq plain len))))))
+
+(defun decode (compressed)
+  (if (zerop (length compressed))
+    ""
+    (let* ((num-string (take-while #'digit-char-p compressed))
```
verdammelt, line 18: Would `parse-integer` work well here rather than `take-while`?

PaulT89: Well, I thought it was a good idea, and so I went back and tried to rewrite it.  Unfortunately there's a problem - `parse-integer` ignores leading spaces instead of treating them as non-digit chars, so:
- `parse-integer " 42sp5q" :junk-allowed t` returns `42 ; 3` instead of `NIL ; 0`
- `parse-integer " hp7q" :junk-allowed t` returns `NIL ; 1` instead of `NIL ; 0`

The weird thing is that it does just fine with trailing spaces:
- `parse-integer "2 5b3z" :junk-allowed t` returns `2 ; 1` 👍 
- `parse-integer "12 " :junk-allowed t` returns `12 ; 2` 👍 

So, yeah, I don't know what's up with that.

PaulT89:
Yeah, I can easily imagine this being remade into a practice exercise for streams.

verdammelt (approved):

PR #602, opened 2022-05-11
Create exercise matching-brackets

PaulT89:
## Summary

Used generator to create Matching Brackets practice exercise.

Difficulty = 2.  Oddly enough, Python and Rust have this as 1, and Ruby has it at 7 - it's definitely not 7, and 1 seems a little low.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/matching-brackets/.meta/example.lisp
+++ b/exercises/practice/matching-brackets/.meta/example.lisp
@@ -4,17 +4,14 @@
 
 (in-package :matching-brackets)
 
+(defparameter +bracket-lookup+ '((#\( . #\)) (#\{ . #\}) (#\[ . #\])))
+
+(defun eliminate-pairs (tracker bracket)
+  (let ((match (rassoc bracket +bracket-lookup+)))
+    (if (and tracker match (char= (car match) (car tracker)))
+      (cdr tracker)
+      (cons bracket tracker))))
+
 (defun pairedp (value)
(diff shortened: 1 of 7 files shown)


verdammelt (commented):
Two comments which are what I'd give as feedback if your example solution were a submitted solution:

The knowledge of brackets and their pairs is spread in two places, I think it would be good form to collect that together. 

Also replace the hard-coding of matching checking with some table lookup type thing perhaps. This makes the code more extensible.

PaulT89:
Good idea.

exercises/practice/matching-brackets/.meta/example.lisp
```diff
@@ -0,0 +1,17 @@
...
+(defparameter +bracket-lookup+ '((#\( . #\)) (#\{ . #\}) (#\[ . #\])))
+
+(defun eliminate-pairs (tracker bracket)
+  (let ((match (rassoc bracket +bracket-lookup+)))
+    (if (and tracker match (char= (car match) (car tracker)))
+      (cdr tracker)
+      (cons bracket tracker))))
+
+(defun pairedp (value)
+  (let ((only-brackets (remove-if-not (lambda (c) (find c "{}[]()")) value)))
```
verdammelt, line 16: Still have the knowledge of brackets split up... maybe generate the the set of all brackets from `+bracket-lookup+`?

PaulT89: Okay.

PR #604, opened 2022-05-11
Create exercise two-bucket

PaulT89:
## Summary

Generated practice exercise Two Bucket.  Still had to do a bit of work to de-clutter the test file (plus add a helper function).

Example is written in Object Oriented style to (hopefully) improve clarity, though knowledge of CLOS is not actually needed to complete this exercise.  Any one (or combination of) a number of normal data structures will suffice (i.e. lists, alists, vectors, or arrays).

Docstring comments may still be required to better explain example.  Feedback is specifically requested regarding this point.

Difficulty = 6.  Other tracks have this at 4 or 5, but I can't help but feel a Common Lisp solution to this exercise requires a little bit more work by comparison.  An OOP-style solution, while easier to keep track of, requires prior knowledge of CLOS (or the desire to find out), while the code in a non-OOP solution can start to look confusing, unless extra care is taken to clearly label all steps of the process (whether through clear function-naming, or commenting, or both).

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/two-bucket/.meta/example.lisp
+++ b/exercises/practice/two-bucket/.meta/example.lisp
@@ -4,36 +4,6 @@
 
 (in-package :two-bucket)
 
-(defun measure (bucket-one bucket-two goal start-bucket)
-  (unless (invalid-input-p bucket-one bucket-two goal)
-    (let* ((first-bucket (make-instance 'bucket :capacity bucket-one :name "one"))
-           (second-bucket (make-instance 'bucket :capacity bucket-two :name "two"))
-           (results (if (string= start-bucket "one")
-                      (iterate-through-puzzle first-bucket second-bucket goal)
-                      (iterate-through-puzzle second-bucket first-bucket goal))))
-      (pairlis '("moves" "goal-bucket" "other-bucket") results))))
-
-(defun invalid-input-p (bucket-one bucket-two goal)
-  (or (plusp (mod goal (gcd bucket-one bucket-two)))
-      (and (> goal bucket-one) (> goal bucket-two))))
-
-(defun iterate-through-puzzle (start-bucket other-bucket goal)
-  (fill-bucket start-bucket)
-  (loop for moves from 1
-        until (= goal (amount start-bucket))
-        until (= goal (amount other-bucket))
-    do (cond
-         ((= goal (capacity other-bucket)) (fill-bucket other-bucket))
-         ((fullp other-bucket) (empty other-bucket))
-         ((zerop (amount start-bucket)) (fill-bucket start-bucket))
-         (t (transfer start-bucket other-bucket)))
-    finally (return (cons moves (retrieve-results start-bucket other-bucket goal)))))
-
-(defun retrieve-results (bucket-one bucket-two goal)
-  (if (= goal (amount bucket-one))
-    (list (name bucket-one) (amount bucket-two))
-    (list (name bucket-two) (amount bucket-one))))
-
 (defclass bucket ()
   ((capacity :initarg :capacity
              :reader capacity)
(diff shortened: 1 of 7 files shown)


exercises/practice/two-bucket/.meta/example.lisp
```diff
@@ -0,0 +1,66 @@
...
+      (pairlis '("moves" "goal-bucket" "other-bucket") results))))
+
+(defun invalid-input-p (bucket-one bucket-two goal)
+  (or (plusp (mod goal (gcd bucket-one bucket-two)))
+      (and (> goal bucket-one) (> goal bucket-two))))
+
+(defun iterate-through-puzzle (start-bucket other-bucket goal)
+  (fill-bucket start-bucket)
+  (loop for moves from 1
+        until (= goal (amount start-bucket))
```
verdammelt, line 23: Why two `until` clauses? Should these simply be made into a single `until` with an `or` expression?

exercises/practice/two-bucket/two-bucket-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+(def-suite* two-bucket-suite)
+
+(defun sort-alist (alist)
+  (sort alist #'string< :key #'first))
+
+(test measure-using-bucket-one-of-size-3-and-bucket-two-of-size-5-start-with-bucket-one
+    (let* ((bucket-one 3)
+           (bucket-two 5)
+           (goal 1)
+           (start-bucket "one")
```
verdammelt, line 25: I think using keywords rather than strings for the bucket names would be more idiomatic.

exercises/practice/two-bucket/two-bucket-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+           (result '(("moves" . 4) ("goal-bucket" . "one") ("other-bucket" . 5)))
+           (output (two-bucket:measure bucket-one bucket-two goal start-bucket)))
```
verdammelt, line 27: Same for the keys of the alist result.

exercises/practice/two-bucket/.meta/example.lisp
```diff
@@ -0,0 +1,66 @@
...
+(defclass bucket ()
+  ((capacity :initarg :capacity
+             :reader capacity)
+   (amount :initform 0
+           :reader amount)
+   (name :initarg :name
+         :reader name)))
+
+(defmethod empty ((obj bucket))
+  (setf (slot-value obj 'amount) 0))
```
verdammelt, line 46: Any reason why these are generic functions rather than ordinary functions?

exercises/practice/two-bucket/.meta/example.lisp
```diff
@@ -0,0 +1,66 @@
...
+
+(defmethod fill-bucket ((obj bucket))
+  (setf (slot-value obj 'amount) (slot-value obj 'capacity)))
```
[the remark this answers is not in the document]
PaulT89: Because `fill` clashes with the `fill` function

exercises/practice/two-bucket/.meta/example.lisp
PaulT89: I thought it was clearer this way, but I can combine them into one if you wish.

exercises/practice/two-bucket/two-bucket-test.lisp
PaulT89: Good point.

exercises/practice/two-bucket/.meta/example.lisp
```diff
@@ -0,0 +1,66 @@
...
+  (= (slot-value obj 'amount) (slot-value obj 'capacity)))
+
+(defmethod overflowingp ((obj bucket))
+  (when (> (slot-value obj 'amount) (slot-value obj 'capacity))
+    (- (slot-value obj 'amount) (slot-value obj 'capacity))))
+
+(defmethod combined-amounts ((bucket-one bucket) (bucket-two bucket))
+  (+ (slot-value bucket-two 'amount) (slot-value bucket-one 'amount)))
+
+(defmethod transfer ((from-bucket bucket) (to-bucket bucket))
```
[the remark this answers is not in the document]
PaulT89: Well, in my mind I imagined the excess water just sort of floating in mid-air above the overflowing bucket - it makes it easier to scoop back up you see 😄 .  Anyway, the way I see it, it makes no difference which way the transfer occurs (pour, test, adjust vs test, adjust, pour) - if the puzzle parameters aren't constrained by real-word practicality, then either way the logic still stacks up.

Then there's the fact that the code looks... well... kind of clunky if I do it the other way (though maybe there's a more elegant way to write it):

````lisp
(defmethod transfer ((from-bucket bucket) (to-bucket bucket))
  (let ((combined (combined-amounts from-bucket to-bucket)))
    (if (> combined (slot-value to-bucket 'capacity))
      (progn
       (setf (slot-value from-bucket 'amount) (- combined (slot-value to-bucket 'capacity)))
       (fill-bucket to-bucket))
      (progn
       (setf (slot-value to-bucket 'amount) combined)
       (empty from-bucket)))))
````
I mean, it works, but I wish it were more streamlined.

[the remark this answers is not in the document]
Comment (PaulT89) [reply exercises/practice/two-bucket/.meta/example.lisp:61]: Well, I say 'the original', though I've since modified it to use `incf` instead of `setf` plus the `combined-amounts` method.

File: exercises/practice/two-bucket/.meta/example.lisp (hunk 4 above)
Comment (PaulT89) [reply exercises/practice/two-bucket/.meta/example.lisp:46]: I wanted to keep everything that 'did stuff' with the bucket(s) somehow linked to the bucket class, and the fact that this is defined as a method (as opposed to your average, run-of-the-mill function), and specifically takes a bucket object/instance best indicates this (or at least I believe so, but I'm up for suggestions).

File: exercises/practice/two-bucket/.meta/example.lisp [hunk 7]
```diff
@@ -0,0 +1,66 @@
...
+         ((zerop (amount start-bucket)) (fill-bucket start-bucket))
+         (t (transfer start-bucket other-bucket)))
+    finally (return (cons moves (retrieve-results start-bucket other-bucket goal)))))
+
+(defun retrieve-results (bucket-one bucket-two goal)
+  (if (= goal (amount bucket-one))
+    (list (name bucket-one) (amount bucket-two))
+    (list (name bucket-two) (amount bucket-one))))
+
+(defclass bucket ()
```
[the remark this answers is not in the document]
PaulT89: Yeah, I'm still kind of new to CLOS - the difference between methods and generics is still kind of a weird thing for me (I'm so used to everything in a class being a method).  Actually, I find the whole idea of generics kind of weird.  Everything that I wanted to be associated with the class I made into a method, but I don't know which of these _exactly_ would be better off as a generic.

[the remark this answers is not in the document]
PaulT89: Oh, and as for the functions, they're not methods because their primary purpose is to solve the puzzle, whereas the methods' primary purpose is to change the state of the bucket(s).  In a nutshell, the functions _utilize_ the buckets, but they don't mess _with_ the buckets - instead they tell the buckets' _methods_ to mess with the buckets.

Honestly, I don't know how orthodox or otherwise this way of doing this is, but it feels right to me.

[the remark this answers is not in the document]
verdammelt: Every method is attached to a generic function. If the implementation cannot find a generic function of the same name it will create one.

Given that no polymorphism is being done all of these *could* be normal functions rather than generic functions & methods.

I'm fine with leaving them as methods - but I think if we do then use `defgeneric` to define the generic functions and the single methods may be better style.

exercises/practice/two-bucket/.meta/example.lisp
verdammelt: In reviewing it again - I think having these generic functions are fine... but see my comment above about how to define generic functions / methods and my understanding of 'good style' with respect to them.

exercises/practice/two-bucket/.meta/example.lisp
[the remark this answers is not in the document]
verdammelt: Oh... would you look at that there is a `fill` function?! 

If you wanted to you could use a shadow declaration in the package to allow you to define a `fill` function... but up to you.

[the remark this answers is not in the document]
PaulT89: Huh, didn't know I could do that.  Cool.

exercises/practice/two-bucket/.meta/example.lisp
```diff
@@ -4,6 +4,8 @@
 
 (in-package :two-bucket)
 
+(shadow 'fill)
```
verdammelt, line 7: 👍 

Also:

```
(defpackage ....
   (:shadow 'fill)
   ....)
```

verdammelt (approved):

PR #607, opened 2022-05-11
Create exercise sum-of-multiples

PaulT89:
## Summary

Generated practice exercise Sum of Multiples.

Difficulty = 3.  Could be a 2 though.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/sum-of-multiples/.meta/example.lisp
+++ b/exercises/practice/sum-of-multiples/.meta/example.lisp
@@ -5,13 +5,6 @@
 (in-package :sum-of-multiples)
 
 (defun sum (factors limit)
- (let ((cleaned-factors (remove-if-not #'plusp factors)))
-   (apply #'+ (remove-duplicates (multiples-of cleaned-factors limit)))))
-
-(defun multiples-of (factors limit)
-  (when factors
-   (let ((factor (first factors))
-         (remaining (rest factors)))
-     (loop for x from factor below limit by factor
-       collect x into output
-       finally (return (append output (multiples-of remaining limit)))))))
+  (loop for factor in (remove-if-not #'plusp factors)
+    collect (loop for x from factor below limit by factor collect x) into output
+    finally (return (apply #'+ (remove-duplicates (reduce #'append output))))))
--- a/exercises/practice/sum-of-multiples/.meta/example.lisp
+++ b/exercises/practice/sum-of-multiples/.meta/example.lisp
@@ -1,0 +1,17 @@
+(defpackage :sum-of-multiples
+  (:use :cl)
+  (:export :sum))
+
+(in-package :sum-of-multiples)
+
+(defun sum (factors limit)
+ (let ((cleaned-factors (remove-if-not #'plusp factors)))
+   (apply #'+ (remove-duplicates (multiples-of cleaned-factors limit)))))
+
+(defun multiples-of (factors limit)
+  (when factors
+   (let ((factor (first factors))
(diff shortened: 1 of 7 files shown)


exercises/practice/sum-of-multiples/.meta/example.lisp
```diff
@@ -0,0 +1,17 @@
...
+
+(in-package :sum-of-multiples)
+
+(defun sum (factors limit)
+ (let ((cleaned-factors (remove-if-not #'plusp factors)))
+   (apply #'+ (remove-duplicates (multiples-of cleaned-factors limit)))))
+
+(defun multiples-of (factors limit)
+  (when factors
+   (let ((factor (first factors))
```
verdammelt, line 13: Should these be put into the `loop`, perhaps with something like `with (factor . remaining) = factors`?

exercises/practice/sum-of-multiples/.meta/example.lisp
```diff
@@ -0,0 +1,17 @@
...
+         (remaining (rest factors)))
+     (loop for x from factor below limit by factor
+       collect x into output
+       finally (return (append output (multiples-of remaining limit)))))))
```
verdammelt, line 17: Why recursion? Why not iterate across the list of factors?

exercises/practice/sum-of-multiples/sum-of-multiples-test.lisp
```diff
@@ -0,0 +1,100 @@
...
+;; Enter the testing package
+(in-package :sum-of-multiples-test)
+
+;; Define and enter a new FiveAM test-suite
+(def-suite* sum-of-multiples-suite)
+
+(test no-multiples-within-limit
+    (let ((factors (list 3 5))
+          (limit 1))
+      (is (equal 0 (sum-of-multiples:sum factors limit)))))
```
verdammelt, line 21: I think we should use `=` in these tests.

verdammelt (commented):
Just some comments. Wondering if we should create some sort of "iteration" concept and put this in it.

exercises/practice/sum-of-multiples/sum-of-multiples-test.lisp
PaulT89: Agreed.

exercises/practice/sum-of-multiples/.meta/example.lisp
PaulT89: Mostly just style choice.  I can rewrite it as a nested loop.

PaulT89:
> Just some comments. Wondering if we should create some sort of "iteration" concept and put this in it.

Isn't iteration slightly different from loops?  I mean, the answer is 'yes' either way, but I'm just curious.

verdammelt:
Iteration (to me) is a concept about how to well.. iterate - maybe through a sequence or via a counter (good old for-loop). Iteration can be implemented in multiple ways - Common Lisp has several built in iteration constructs but are generally `do` (and its kin) and `loop` and I guess recursion.

So maybe we need a concept of iteration with branches for `do` and `loop` (where each of those have sub-branches based upon functionality (i.e. `do times` vs `dolist` vs all the crazy `loop` `for` keywords...)

PaulT89:
Okay.  I remember why I thought the term 'iterate' meant something else - I kept seeing it in [the loops section of the Common Lisp Cookbook](https://lispcookbook.github.io/cl-cookbook/iteration.html) and didn't realize that it's just another loop macro, just not a built-in one.

PR #609, opened 2022-05-12
Create exercise isbn-verifier

PaulT89:
## Summary

Generated practice exercise Isbn Verifier.

Difficulty = 5.  This is the same kind of exercise as Luhn, though there does seem to be a bit more checking required to get this right.  Could be raised to a 6.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/isbn-verifier/.meta/example.lisp
+++ b/exercises/practice/isbn-verifier/.meta/example.lisp
@@ -11,6 +11,6 @@
   (let ((cleaned (remove-if-not #'alphanumericp isbn)))
     (when (= 10 (length cleaned)) ; Check isbn is 10 chars long
       (let ((num-list (map 'list #'char-to-num cleaned))) ; Map chars to numbers (or nil)
-        (unless (some #'null num-list) ; Return nil if any nils in num-list
-          (unless (member 10 (butlast num-list)) ; Return nil if any of the first 9 nums eql 10
-            (zerop (mod (apply #'+ (mapcar #'* num-list '(10 9 8 7 6 5 4 3 2 1))) 11))))))))
+        ; Return nil if any nils in num-list or if any of the first 9 nums eql 10
+        (unless (or (some #'null num-list) (member 10 (butlast num-list)))
+          (zerop (mod (apply #'+ (mapcar #'* num-list '(10 9 8 7 6 5 4 3 2 1))) 11)))))))
--- a/exercises/practice/isbn-verifier/.meta/example.lisp
+++ b/exercises/practice/isbn-verifier/.meta/example.lisp
@@ -1,0 +1,17 @@
+(defpackage :isbn-verifier
+  (:use :cl)
+  (:export :validp))
+
+(in-package :isbn-verifier)
+
+(defun validp (isbn)
+  (let ((cleaned (remove-if-not #'alphanumericp isbn)))
+    (when (= 10 (length cleaned)) ; Check isbn is 10 chars long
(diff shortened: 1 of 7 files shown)


exercises/practice/isbn-verifier/.meta/example.lisp
```diff
@@ -0,0 +1,17 @@
...
+
+(in-package :isbn-verifier)
+
+(defun validp (isbn)
+  (let ((cleaned (remove-if-not #'alphanumericp isbn)))
+    (when (= 10 (length cleaned)) ; Check isbn is 10 chars long
+      (let ((num-list (map 'list #'char-to-num cleaned))) ; Map chars to numbers (or nil)
+        (unless (some #'null num-list) ; Return nil if any nils in num-list
+          (unless (member 10 (butlast num-list)) ; Return nil if any of the first 9 nums eql 10
+            (setq total (apply #'+ (mapcar #'* num-list '(10 9 8 7 6 5 4 3 2 1))))
```
verdammelt, line 13: One should not use `setq` on an undefined symbol. I would use `let` here instead - or not even bother with a new symbol.

verdammelt (commented):
I think this looks good. If you could make it so that the example file loaded without warnings then we can merge.

The warnings I'm seeing in SBCL:

```
CL-USER> (load ".meta/example.lisp")

; file: /Users/mark/SRC/exercism/common-lisp/exercises/practice/isbn-verifier/.meta/example.lisp
; in: DEFUN VALIDP
;     #'ISBN-VERIFIER::CHAR-TO-NUM
; 
; caught STYLE-WARNING:
;   undefined function: ISBN-VERIFIER::CHAR-TO-NUM

;     (MOD ISBN-VERIFIER::TOTAL 11)
; 
; caught WARNING:
;   undefined variable: ISBN-VERIFIER::TOTAL
; 
; compilation unit finished
;   Undefined function:
;     CHAR-TO-NUM
;   Undefined variable:
;     TOTAL
;   caught 1 WARNING condition
;   caught 1 STYLE-WARNING condition
WARNING: redefining ISBN-VERIFIER:VALIDP in DEFUN
T
```

exercises/practice/isbn-verifier/.meta/example.lisp
PaulT89: Okay.

PaulT89:
Also, now that I think about it, do you want me to put the two `unless`'s together into a single `unless (or` or is it better to keep it the way it is (i.e. with one quick comment for each)?

exercises/practice/isbn-verifier/.meta/example.lisp
```diff
@@ -0,0 +1,16 @@
...
+(defun char-to-num (input-char)
+  (if (char= #\X input-char) 10 (digit-char-p input-char)))
+
+(defun validp (isbn)
+  (let ((cleaned (remove-if-not #'alphanumericp isbn)))
+    (when (= 10 (length cleaned)) ; Check isbn is 10 chars long
+      (let ((num-list (map 'list #'char-to-num cleaned))) ; Map chars to numbers (or nil)
+        (unless (some #'null num-list) ; Return nil if any nils in num-list
+          (unless (member 10 (butlast num-list)) ; Return nil if any of the first 9 nums eql 10
+            (zerop (mod (apply #'+ (mapcar #'* num-list '(10 9 8 7 6 5 4 3 2 1))) 11))))))))
```
verdammelt, line 16: I think a single conditional expression makes sense here rather than nested ones.

PaulT89: Will do.

PR #610, opened 2022-05-13
Create exercise queen-attack

PaulT89:
## Summary

Generated practice exercise Queen Attack.

Difficulty = 6.  On the face of it, this exercise seems like nothing more than a simple logic problem, however, it requires the use of structs, which may be a challenge to the uninitiated.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/queen-attack/queen-attack-test.lisp
+++ b/exercises/practice/queen-attack/queen-attack-test.lisp
@@ -16,63 +16,63 @@
 (def-suite* queen-attack-suite)
 
 (test queen-with-a-valid-position
-    (let ((coordinates '((:row . 2) (:column . 2))))
-      (is (equalp #S(queen-attack:queen :row 2 :column 2) (queen-attack:create coordinates)))))
+    (let ((coordinates '(2 . 2)))
+      (is-true (queen-attack:valid-position-p coordinates))))
 
 (test queen-must-have-positive-row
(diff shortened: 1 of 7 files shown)


verdammelt (commented):
You don't *need* a struct. You can simply use a cons and then define `queen-row` and `queen-column` yourself.

Seems like this is not such a complex exercise - so 6 seems too high to me.

exercises/practice/queen-attack/queen-attack-test.lisp
```diff
@@ -0,0 +1,80 @@
...
+
+;; Enter the testing package
+(in-package :queen-attack-test)
+
+;; Define and enter a new FiveAM test-suite
+(def-suite* queen-attack-suite)
+
+(test queen-with-a-valid-position
+    (let ((coordinates '((:row . 2) (:column . 2))))
+      (is (equalp #S(queen-attack:queen :row 2 :column 2) (queen-attack:create coordinates)))))
```
verdammelt, line 20: This forces the use of a struct as the implementation. Perhaps we can have the two accessors also be exported which will allow the student to use any implementation they wish?

PaulT89:
Okay, I've completely redone the exercise.  It doesn't use structs anymore, a coordinate pairing is just a cons, validation is just a function, difficulty is now 2.

verdammelt (approved):

PR #613, opened 2022-05-13
Create exercise knapsack

PaulT89:
## Summary

Generated practice exercise Knapsack.

Difficulty = 7.  This is a genuinely tricky exercise.  It will either take little code and an awful lot of thought, or a lot of code and a reasonable amount of thought.

## Checklist
- [x] CI is green


exercises/practice/knapsack/.meta/example.lisp
```diff
@@ -0,0 +1,21 @@
...
+(in-package :knapsack)
+
+(defun maximum-value (maximum-weight items)
+  (let ((compact-items (loop for item in items
+                         collect (cons (cdr (assoc :weight item))
+                                       (cdr (assoc :value item)))))
+        (possible-combos '((0 . 0))))
+    (loop for item in compact-items
+      do (setf possible-combos (combine-item item possible-combos)))
+    (delete-if (lambda (x) (> x maximum-weight)) possible-combos :key #'car)
```
verdammelt, line 14: Careful here - `delete-if` may or may not modify the sequence. One should use `setf` here to rebind `possible-combos` and perhaps simply use `remove-if` as well (unless worried about the possibility of making too much garbage during the run - but then i'd want to use `room` to check that...)

exercises/practice/knapsack/knapsack-test.lisp
```diff
@@ -0,0 +1,87 @@
...
+(def-suite* knapsack-suite)
+
+(test no-items
+    (let ((maximum-weight 100)
+          (items (list)))
+      (is (= 0 (knapsack:maximum-value maximum-weight items)))))
+
+(test one-item-too-heavy
+    (let ((maximum-weight 10)
+          (items (list (list '(:weight . 100) '(:value . 1)))))
```
verdammelt, line 25: Why not just quote the whole thing rather than two layers of lists and then quoted items?

exercises/practice/knapsack/.meta/example.lisp
PaulT89: I don't know.  I thought the Hyperspec implied that the only way a `delete-if` won't modify the sequence is if the predicate never returns true.  It even explicitly states:

> delete, delete-if, and delete-if-not return a sequence of the same type as sequence that has the same elements except that those in the subsequence bounded by _start_ and _end_ and _satisfying the test_ have been deleted.

I mean, I suppose there could be some weird code under the hood which, under some circumstances, returns an unmodified sequence when it ought to return a modified sequence, but I can't begin to fathom what those circumstances would be.

exercises/practice/knapsack/knapsack-test.lisp
PaulT89: It's generated that way to avoid problems with nesting (i.e. quoted lists within quoted lists).  I can clean this up though.

exercises/practice/knapsack/.meta/example.lisp
PaulT89: Okay.  But what's the point of delete if it might not modify the sequence?

verdammelt: My understanding is that either (or both) 1) backward compatibility with other Lisps which had `delete` 2) allows implementations to use a destructive algorithm if they choose.

`remove` is never destructive, `delete` is destructive if the implementation wants/can.

verdammelt (approved):

PR #615, opened 2022-05-14
Create exercise rail-fence-cipher

PaulT89:
## Summary

Generated practice exercise Rail Fence Cipher.

Difficulty = 4.  Hovers around here for other tracks.  Seems good enough.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/rail-fence-cipher/.meta/example.lisp
+++ b/exercises/practice/rail-fence-cipher/.meta/example.lisp
@@ -5,15 +5,6 @@
 
 (in-package :rail-fence-cipher)
 
-(defun encode (msg rails)
-  (let ((fence (make-fence (length msg) rails)))
-    (populate-and-sort msg fence #'post-row)))
-
-(defun decode (msg rails)
-  (let ((fence (make-fence (length msg) rails)))
-    (sort fence #'< :key #'post-row) ; rearrange the fence into its encoded state
-    (populate-and-sort msg fence #'post-column)))
-
 (defstruct post letter row column)
 
 (defun make-fence (seq-length rails)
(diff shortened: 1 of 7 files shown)


exercises/practice/rail-fence-cipher/.meta/example.lisp
```diff
@@ -0,0 +1,36 @@
...
+
+(defstruct post letter row column)
+
+(defun make-fence (seq-length rails)
+  "Function make-fence will output a list of structs of type post, which will come
+to represent a character and its coordinates in the fence.  Function contains a
+loop that goes along the columns, and uses mathematical equations to find the
+correct row.  Letters are not filled in here, and are set to nil."
+  (loop for i from 0 below seq-length with r = (1- rails)
+    if (< (mod i (* 2 r)) r) ; if i mod 2r < r then...
```
verdammelt, line 25: I suggest removing these comments - I don't think they are actually helping are they?

exercises/practice/rail-fence-cipher/.meta/example.lisp
```diff
@@ -0,0 +1,36 @@
...
+    collect (make-post :letter nil :column i :row (mod i r)) ; row = i mod r, else...
+    else
+    collect (make-post :letter nil :column i :row (- r (mod i r))))) ; row = r - (i mod r)
+
+(defun populate-and-sort (msg fence sort-key)
```
verdammelt, line 30: Feels like this function is doing too much, it populates the fence, sorts the fence and then unpacks the fence.

Maybe if it was going to do all of that maybe it should just make the fence as well? (yes i know that won't work in the way you wrote decode).

I wonder if maybe populate-and-sort should do less? Maybe unpacking a fence is the responsibility of some other function? Perhaps we need 4 functions in the fence interface: a constructor, a writer (string->fence), a sorter (fence -> fence), and a reader (fence -> string).

exercises/practice/rail-fence-cipher/.meta/example.lisp
PaulT89: For me they make it clear at a glance - but that's just me.  I have trouble with prefix notation arithmetic.

exercises/practice/rail-fence-cipher/.meta/example.lisp
PaulT89: The reason it does so much is mostly just to decrease duplication, but I can split it into 3 separate functions if you wish.

verdammelt (approved):

PR #616, opened 2022-05-15
Create exercise pig-latin

PaulT89:
## Summary

Generated practice exercise Pig Latin.

Difficulty = 6.  This is higher than other tracks.  Ruby and Rust, for example, have this set at 4, however these languages have lots of string helper functions that Common Lisp doesn't have (e.g. split, join, starts with, etc.).  The extra difficulty comes from having to make your own helper functions.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/pig-latin/.meta/example.lisp
+++ b/exercises/practice/pig-latin/.meta/example.lisp
@@ -6,23 +6,6 @@
 
 (defparameter +vowels+ '("a" "e" "i" "o" "u" "xr" "yt"))
 
-(defun translate (phrase)
-  (loop for word in (split-string phrase)
-    collect (translate-word word) into output
-    finally (return (format nil "~{~A~^ ~}" output))))
-
-(defun translate-word (word)
-  (let ((index (take-consonants word)))
-    (format nil "~A~Aay" (subseq word index) (subseq word 0 index))))
-
-(defun take-consonants (word &optional (index 0))
-  (let ((stopp (lambda (w i) (or (starts-with-p w +vowels+)
-                                 (and (char= #\y (char w 0)) (= 1 i))))))
-    (cond
-      ((funcall stopp word index) index)
-      ((starts-with-p word '("qu")) (take-consonants (subseq word 2) (+ 2 index)))
-      (t (take-consonants (subseq word 1) (1+ index))))))
-
 (defun split-string (str)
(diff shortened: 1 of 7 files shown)


exercises/practice/pig-latin/.meta/example.lisp
```diff
@@ -0,0 +1,37 @@
...
+  (loop for word in (split-string phrase)
+    collect (translate-word word) into output
+    finally (return (format nil "~{~A~^ ~}" output))))
+
+(defun translate-word (word)
+  (let ((index (take-consonants word)))
+    (format nil "~A~Aay" (subseq word index) (subseq word 0 index))))
+
+(defun take-consonants (word &optional (index 0))
+  (let ((stopp (lambda (w i) (or (starts-with-p w +vowels+)
```
verdammelt, line 19: Maybe this is a case where `stop-p` might be better? `stopp` looks like a typo :)

exercises/practice/pig-latin/.meta/example.lisp
```diff
@@ -0,0 +1,37 @@
...
+  (unless (zerop (length str))
+    (let* ((split-at (position #\space str))
+           (word (subseq str 0 split-at))
+           (rest-str (string-left-trim " " (subseq str (length word)))))
+      (cons word (split-string rest-str)))))
+
+(defun starts-with-p (word search-list)
+  (when search-list
+    (case (search (car search-list) word)
+      (0 T)
```
verdammelt, line 36: Maybe instead of recursion you could create a expression with `any` and a predicate function that sees if an item from search-list is the first thing in word?

exercises/practice/pig-latin/.meta/example.lisp
PaulT89: Agreed.  And now that I think about it, I'll also just make this an actual function instead of a named lambda.

exercises/practice/pig-latin/.meta/example.lisp
PaulT89: I came up with the following, but it doesn't look particularly neat:

````lisp
(defun starts-with-p (word search-list)
  (some (lambda (str) (when (setf res (search str word)) (zerop res))) search-list))
````

Is it really an improvement?

verdammelt (approved):

PR #617, opened 2022-05-15
Create exercise flatten-array

PaulT89:
## Summary

Generated practice exercise Flatten Array.

Difficulty = 7.  I don't know if I'm just tired or something, but this deceptively easy-looking exercise was nightmarishly difficult for me to figure out.

## Checklist
- [x] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -696,6 +696,15 @@
         "practices": ["conditionals", "functions"],
         "prerequisites": ["conditionals", "functions"],
         "status": "beta"
+      },
+      {
+        "slug": "flatten-array",
+        "name": "Flatten Array",
+        "uuid": "008ae436-b9d5-4882-b1c7-8809baff70ea",
+        "difficulty": 7,
+        "practices": ["functions", "conditionals"],
+        "prerequisites": ["functions", "lists", "conditionals"],
+        "status": "beta"
       }
     ],
     "foregone": [
--- a/config.json
+++ b/config.json
@@ -701,9 +701,9 @@
         "slug": "flatten-array",
         "name": "Flatten Array",
         "uuid": "008ae436-b9d5-4882-b1c7-8809baff70ea",
-        "difficulty": 7,
-        "practices": ["functions", "conditionals"],
-        "prerequisites": ["functions", "lists", "conditionals"],
+        "difficulty": 6,
+        "practices": ["reducing"],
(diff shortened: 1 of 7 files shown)


verdammelt (commented):
I don't think 7 is the right difficulty level (but I'm saying that without having tried it myself! :) ).

I think a first-pass, inefficient algorithm is something like:

```
(while (notevery #'atom list)
  (setf list (apply #'append list)))
```

right?

PaulT89:
Didn't work for me.  Append doesn't like encountering atoms.

verdammelt:
Below is my attempt at this exercise. While my `apply append` idea would not work as you pointed out I think this is pretty simple. (It does ignore the issue of circular lists but I don't think that is something we need to handle for this exercise...)

```
(defun flatten (nested)
  (remove nil
          (reduce #'(lambda (acc item)
                      (append acc (if (atom item) (list item) (flatten item))))
                  nested :initial-value (list))))
```

Here I used `reduce` but could also do it entirely recursively like so:

```
(defun flatten (nested)
  (labels ((%flatten (nested result)
             (let ((head (first nested))
                   (rest (rest nested)))
               (cond ((endp nested) result)
                     ((null head) (%flatten rest result))
                     (t (%flatten rest
                                  (append result
                                          (if (atom head)
                                              (list head)
                                              (%flatten head (list))))))))))
    (%flatten nested (list))))
```

PaulT89:
I do like the `reduce` version.  I also figured out how to do it the style you mentioned last week (using `while (notevery '#atom` and `apply #'append`):

````lisp
(defun flatten (nested)
  (loop while (notevery #'atom nested)
    do (setf nested (apply #'append (loop for item in nested
                                      collect (if (atom item) (list item) item))))
    finally (return (remove nil nested))))
````

And now that I'm comparing the two, this is just a loopy version of what you did with `reduce` (except with side-effects, though `copy-list` would fix that in a heartbeat).  So, yeah, I think using your `reduce` version would be the better choice, as it seems more Lispy, and it doesn't have any side-effects (not that it matters in this test).

If a difficulty of 7 is still too high, then, given how much thought we both had to put into this, maybe it should only be dropped by 1 level (to 6).

verdammelt:
Thanks for taking the time on this one. I think it looks good so I will merge it!

PR #618, opened 2022-05-16
Create exercise protein-translation

PaulT89:
## Summary

Generated practice exercise Protein Translation.

Difficulty = 5.  Relatively simple exercise with multiple ways of completing it.  The reason that the difficulty is a 5 (and not a 3 or a 4) is that that the student will need to find out about error signalling (like in the Nucleotide Count exercise).

## Checklist
- [x] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -570,6 +570,15 @@
         "practices": ["lists", "mapping", "functions"],
         "prerequisites": ["conditionals", "lists", "mapping", "arithmetic", "functions"],
         "status": "beta"
+      },
+      {
+        "slug": "protein-translation",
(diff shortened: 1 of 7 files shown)


verdammelt (commented):
I like this... before I saw your solution I was thinking of using an alist and strings as keys but I like your use of case with keywords.

Why don't we set its difficulty to 4 like nucleotide-count?

Can you fix the example file so it doesn't have warnings on load? Then I suggest we merge it.

PaulT89:
Fixed.

verdammelt (approved):

PR #620, opened 2022-05-18
Create exercise matrix

PaulT89:
## Summary

Generated practice exercise Matrix.

Difficulty = 1.  I altered the tests for this exercise - it used to have strings as input (which the student would then have to parse) but that made the test file look weird.  Instead, I made this an arrays-only practice exercise (with the use of loops).

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/matrix/.docs/instructions.md
+++ b/exercises/practice/matrix/.docs/instructions.md
@@ -1,0 +1,41 @@
+# Instructions
+
+Given a string representing a matrix of numbers, return the rows and columns of
+that matrix.
+
+So given a string with embedded newlines like:
+
+```text
+9 8 7
+5 3 2
+6 6 7
+```
+
+representing this matrix:
+
+```text
+    1  2  3
+  |---------
+1 | 9  8  7
+2 | 5  3  2
+3 | 6  6  7
+```
+
+your code should be able to spit out:
+
+- A list of the rows, reading each row left-to-right while moving
+  top-to-bottom across the rows,
+- A list of the columns, reading each column top-to-bottom while moving
+  from left-to-right.
+
+The rows for our example matrix:
+
+- 9, 8, 7
+- 5, 3, 2
+- 6, 6, 7
+
+And its columns:
+
+- 9, 5, 6
+- 8, 3, 6
+- 7, 2, 7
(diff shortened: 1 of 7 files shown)


exercises/practice/matrix/.docs/instructions.md
```diff
@@ -0,0 +1,41 @@
+# Instructions
+
+Given a string representing a matrix of numbers, return the rows and columns of
```
verdammelt, line 3: Since this no longer takes strings these instructions will need to be modified.

verdammelt (commented):
How bad would this exercise be if it included the parsing of the input strings?

Student would need to write a split function that allowed splitting on newline or space. Then creating a 2D array and populating it. Then writing the row and column functions you already have.

The string splitting is probably the hardest - but by this point likely the student has that function in their repertoire...

PaulT89:
> How bad would this exercise be if it included the parsing of the input strings?
> 
> Student would need to write a split function that allowed splitting on newline or space. Then creating a 2D array and populating it. Then writing the row and column functions you already have.
> 
> The string splitting is probably the hardest - but by this point likely the student has that function in their repertoire...

Yeah, the problem isn't the parsing of strings, the problem is the aesthetics of the test file - inputting a matrix as a string means that the matrix becomes unindented and therefore misaligned from the first newline.  It makes it hard to read and messes up the indentation of the test file.

I suppose I could go back and rewrite the test so that there's a helper function that converts the 2d array into a string as it's being fed into the input.  It would solve the 'ugliness' problem, but on the other hand, it wouldn't necessarily be clear to the student that they're dealing with string input, instead of array input.

Another option would perhaps be to write each line as a separate string as part of a list and then weld the bits together.  For example:

````lisp
(let ((matrix (format nil "~{~A~&~}" '("1 2 3"
                                       "4 5 6"
                                       "7 8 9"))))
  ...
````

Would that be better?

exercises/practice/matrix/.docs/instructions.md
PaulT89: Maybe, maybe not - I have new thoughts.  See conversation.

PaulT89:
Okay, I've completely overhauled the exercise - i.e rewritten the example, rewritten the test file, renamed variables, changed the difficulty, etc.

verdammelt (approved):
I think i like how it is now. approving. Thanks.

PR #621, opened 2022-05-22
Create exercise spiral-matrix

PaulT89:
## Summary

Generated practice exercise Spiral Matrix.

Difficulty = 3.  Rewards those who figure out that lists can be circular.  This is very much an exercise where the amount of code used to reach a solution is inversely proportional to the amount of thought put in.  Still eminently doable in any case.

## Checklist
- [x] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -696,6 +696,15 @@
         "practices": ["conditionals", "functions"],
         "prerequisites": ["conditionals", "functions"],
         "status": "beta"
+      },
+      {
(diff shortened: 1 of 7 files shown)


verdammelt (approved):

verdammelt:
I have a little concern about the circular lists. They are mentioned in the lists concept but perhaps they'll want their own concept some day. I'll merge this in for now however.

PR #624, opened 2022-05-24
Create exercise say

PaulT89:
## Summary

Generated practice exercise Say.

Difficulty = 1.  Quite an involved exercise in other languages... but hilariously easy in Lisp!

## Checklist
- [x] CI is green


verdammelt:
Oh fun - another one that is handled by `format`! I think we'll want something like `format-numbers` concept to cover `~R` and family.

PR #625, opened 2022-05-25
Create exercise diamond

PaulT89:
## Summary

Generated practice exercise Diamond.

Difficulty = 5.  The amount of code needed to complete this exercise will be inversely proportional to the student's knowledge of all the fancy stuff that the format function can do.  If there is ever a format-advanced concept, this should probably go there, and with a lower difficulty.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/diamond/.meta/example.lisp
+++ b/exercises/practice/diamond/.meta/example.lisp
@@ -1,0 +1,14 @@
+(defpackage :diamond
+  (:use :cl)
+  (:export :rows))
+
+(in-package :diamond)
+
+(defun rows (letter)
+  (let* ((letters (loop for i from 65 to (char-code letter) collect (code-char i)))
+         (columns (1- (* 2 (length letters))))
+         (unique-rows (loop for c in letters and idx from 0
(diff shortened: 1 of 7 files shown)


exercises/practice/diamond/.meta/example.lisp
```diff
@@ -0,0 +1,14 @@
+(defpackage :diamond
+  (:use :cl)
+  (:export :rows))
+
+(in-package :diamond)
+
+(defun rows (letter)
+  (let* ((letters (loop for i from 65 to (char-code letter) collect (code-char i)))
```
verdammelt, line 8: I don't think I want to worry about it right now but if you review:

http://www.lispworks.com/documentation/HyperSpec/Body/13_af.htm

You can see fourth bullet point that points out that characters are not *necessarily* contiguous. So if `#\A` has a `code-char` of 1 `#\B` is not guaranteed to have a `code-char` of 2.

Further your `65` here assumes ASCII. Which is *very* safe - but keeps this code from being truly conforming.

As I said, let's not worry about that here. But wanted to let you know about this annoying little corner of the specification.

verdammelt (approved):

PR #627, opened 2022-05-25
Create exercise affine-cipher

PaulT89:
## Summary

Generated practice exercise Affine Cipher.

Difficulty = 4.  This is about where it's at for other tracks.

## Checklist
- [x] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -687,6 +687,15 @@
         "practices": ["strings", "arrays"],
         "prerequisites": ["strings", "arrays", "functions", "lists"],
         "status": "beta"
+      },
+      {
+        "slug": "affine-cipher",
+        "name": "Affine Cipher",
+        "uuid": "b7885c80-d4d9-4636-a9c2-6051c6f9204c",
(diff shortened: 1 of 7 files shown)


verdammelt:
For a previous exercise I said let's not worry about the assumptions of ASCII etc. But perhaps for this one we should. Is there any reason why this exercise's solution should be assuming ASCII? There are a smattering of magic numbers in the code `97`, `26`, `25`. The latter two I think are valid - they are the length of the Roman alphabet. But `97` is the index of `a` in the ASCII table. 

I suggest that the Char->Int and Int->Char computations are done simply on a lookup of that letter in the Roman alphabet, thus removing any concern for what encoding system the implementation uses.

Discussion of the assumptions of ASCII can be part of the mentoring sessions.

verdammelt (changes requested):
I think that this exercise should not assume ASCII - or any other encoding.

verdammelt (approved):

PR #632, opened 2022-05-25
Create exercise reverse-string

PaulT89:
## Summary

Generated practice exercise Reverse String.

Difficulty = 1.  Honestly, this is so trivially easy that I'm wondering if this exercise should be added to the 'foregone' list.

## Checklist
- [x] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -687,6 +687,15 @@
         "practices": ["strings", "arrays"],
         "prerequisites": ["strings", "arrays", "functions", "lists"],
         "status": "beta"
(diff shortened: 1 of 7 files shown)


verdammelt:
This does look like a case where we should add it to forgone. Unlike roman-numeral where the simple trick is interesting - this is not really interesting.

Instead something like this should be added to the string/sequence concepts showing that strings are sequences and thus sequence operations like reverse work fine on strings.

PR #638, opened 2022-05-27
Create exercise pythagorean-triplet

PaulT89:
## Summary

Generated practice exercise Pythagorean Triplet.

Difficulty = 6.  This is around what other tracks have this set at.  Rust has it 7 and Ruby at 5, though, oddly enough, I found it harder to do in Ruby than in Rust.

## Checklist
- [ ] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -687,6 +687,15 @@
         "practices": ["strings", "arrays"],
         "prerequisites": ["strings", "arrays", "functions", "lists"],
         "status": "beta"
+      },
+      {
+        "slug": "pythagorean-triplet",
+        "name": "Pythagorean Triplet",
+        "uuid": "f7bbf16d-77d9-4d13-a20c-e60f416de764",
+        "difficulty": 6,
+        "practices": ["arithmetic", "lists"],
+        "prerequisites": ["arithmetic", "lists"],
+        "status": "beta"
       }
     ],
     "foregone": ["accumulate", "bank-account", "list-ops", "linked-list", "simple-linked-list"]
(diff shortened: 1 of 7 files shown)


verdammelt (changes requested):
I think this example needs some documentation... is this a well known algorithm? At least a link to a reference of that algorithm. Thanks.

PaulT89:
I didn't really get it from anywhere - it's mostly just pen, paper, and math.  You're right, though - I should probably explain how the whole thing works.  Should I do that at the beginning of the file with `;;;;`s?

verdammelt:
> I didn't really get it from anywhere - it's mostly just pen, paper, and math. You're right, though - I should probably explain how the whole thing works. Should I do that at the beginning of the file with `;;;;`s?

I think a doc string (e.g. "this function uses the Smith-Wombat-Smythe algorithm as described at www.wikipedia...") or other comment at the function level may be sufficient in this case.

verdammelt:
> it's mostly just pen, paper, and math

I understand the first too... it is the last one that trips me up :D

PaulT89:
I just wrote the math and the algorithm out in their entirety.  I can't read the math stuff on wikipedia - that all goes way over my head (due to all the unfamiliar symbols and terminology).  The only thing close to what I did (that I found) was this [GeeksForGeeks](https://www.geeksforgeeks.org/pythagorean-triplet-with-given-sum-using-single-loop/) page, but it only finds the first instance of a triplet, it over-searches, it uses a slightly different equation to find for b, then c, rather than the other way around, etc.  

I don't know - it's really hard to find an algorithm that matches something that I just made up in my head (even though, statistically speaking, it _must_ exist somewhere else).

PaulT89:
And yes, I was lazy and commented using block quotes.

PR #641, opened 2022-05-27
Remove assumptions about characters.

verdammelt:
The Common Lisp specification does not specify the exact ordering of
characters. So we should not assume ASCII or that all
upper/lower/digit characters are contiguous.

-------

## Summary

(Explain what this change is attempting to accomplish)

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/atbash-cipher/.meta/config.json
+++ b/exercises/practice/atbash-cipher/.meta/config.json
@@ -1,6 +1,7 @@
 {
   "authors": [
-    "verdammelt"
+    "verdammelt",
+    "PaulT89"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/atbash-cipher/.meta/example.lisp
+++ b/exercises/practice/atbash-cipher/.meta/example.lisp
@@ -15,23 +15,13 @@
                (zerop (mod i group-size)))
      collect #\Space))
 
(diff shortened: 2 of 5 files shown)


verdammelt:
@PaulT89 - Could you give these a review? After commenting on your PRs I wanted to make sure i wasn't being too hypocritical :)

PaulT89:
**Atbash Cipher**
Honestly, I'd get rid of everything except for the `group` function (and the `encode` function, obviously 😄).  Since the Atbash Cipher only has 1 possible key, create an alist at the beginning:

````lisp
(defparameter +key+ (pairlis 
                     (coerce "abcdefghijklmnopqrstuvwxyz1234567890" 'list)
                     (coerce "zyxwvutsrqponmlkjihgfedcba1234567890" 'list)))
````

...and now, `encode` is just a matter of filtering, mapping, and grouping.

````lisp
(defun encode (plaintext)
  (let* ((filtered (remove-if-not #'alphanumericp (string-downcase plaintext)))
         (ciphertext (map 'string (lambda (c) (cdr (assoc c +key+))) filtered)))
        (format nil "~{~A~^ ~}" (group ciphertext)))) 
````

Easy peezy!

verdammelt:
Oh I like your idea for atbash - will do that!

PR #642, opened 2022-05-28
Create exercise palindrome products

PaulT89:
## Summary

Generated practice exercise Palindrome Products.

Difficulty = 6.  This exercise may be somewhat restrictive in how it needs to be done, as (insufficient) speed and memory may be an issue.

## Checklist
- [x] CI is green


verdammelt:
What is up with the two example files? Can we remove one of them?

PaulT89:
> What is up with the two example files? Can we remove one of them?

Sure.  I only included the second because it looks more elegant and Lispy than the first (I think), but the first is faster and won't risk overflowing the heap.

PR #648, opened 2022-06-05
Update instructions.md

dsletten:
The value 365.25 days/year represents an average Julian calendar year. The correct value for the Gregorian calendar is 365.2425.

## Checklist
- [ ] CI is green


verdammelt:
Thank you for this PR but this text is taken from the official problem-specifications repository for this exercise. So it should be changed there and then synced to this (and other) tracks.

PR #649, opened 2022-06-05
Update space-age-test.lisp

dsletten:
The value for Mercury is wrong for Gregorian calendar years (see https://github.com/exercism/common-lisp/pull/648)

## Checklist
- [ ] CI is green


verdammelt:
Thanks for the PR but this data is taken from the problem-specifications repository (https://github.com/exercism/problem-specifications) and so this change should be submitted there.

Thanks.

Issue #450, opened 2021-08-28
Meta-Issue: review practice exercises for correct concepts.


verdammelt:
Need to go through practice exercises again to ensure they have appropriate concepts in practices and prerequisites. 

Likely plenty of concepts will need to be created as well.

PR #652, opened 2022-06-07
Attempt to fix Issue #640

### LINKED ISSUE #640: Some concepts used too many times
From `./bin/configlet lint`

``` text
Warning: The Concept `strings` appears 16 times in the `practices` arrays of user-facing Practice Exercises, but can only appear at most 10 times:
/Users/mark/SRC/exercism/common-lisp/config.json

Warning: The Concept `arithmetic` appears 16 times in the `practices` arrays of user-facing Practice Exercises, but can only appear at most 10 times:
/Users/mark/SRC/exercism/common-lisp/config.json

Warning: The Concept `conditionals` appears 15 times in the `practices` arrays of user-facing Practice Exercises, but can only appear at most 10 times:
/Users/mark/SRC/exercism/common-lisp/config.json
```

We'll need to check if exercises using these concepts are *practicing* these concepts or just *using* them. Not every exercise that uses strings for example needs to list `strings` as a concept it practices.

PaulT89:
## Summary

First attempt to fix Issue #640.

## Checklist
- [x] CI is green

### DIFF
--- a/config.json
+++ b/config.json
@@ -207,7 +207,7 @@
         "name": "Rna Transcription",
         "uuid": "5c215cbf-a575-42ad-9b3f-892410a63077",
         "difficulty": 1,
-        "practices": ["characters", "mapping"],
+        "practices": ["mapping"],
         "prerequisites": ["functions", "characters", "mapping"],
         "status": "beta"
       },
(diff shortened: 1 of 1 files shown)


verdammelt:
@PaulT89 thanks for the time you put into this but I think it would be better to make plan to tackle #450 instead of just trying to attack the problem of a some exercises having "too many" concepts.

PR #669, opened 2022-10-07
Added asd systems and command.org

informatimago:
## Summary

Added ASDF systems, and a command.org file.

### DIFF
--- a/exercises/practice/hello-world/commands.org
+++ b/exercises/practice/hello-world/commands.org
@@ -1,0 +1,44 @@
+;;;; -*- mode:org; coding:utf-8 -*-
+
+* Testing hello-world
+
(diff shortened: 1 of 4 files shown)


verdammelt:
Thank you for this submission. However I do not think I will merge it. 

For the exercises the track just uses a simple `load` of the "production code" from the "test code" rather than creating ASDF systems. We do use ASDF systems for code related to the track such as the CI system, the test runner, representer and analyzer, along with utilities. If we were to use ASDF systems for the exercises we'd want to implement it across all exercises, update the instructions for the user downloaded to their machine if they use the CLI, use it in our CI, and also use it in our test runner to make the running of the exercise similar between in-browser and on the learner's machine.

Thank you again.

PR #687, opened 2022-11-26
Tests should not check for exact equality of floats

### LINKED ISSUE #686: Pizza Pier test failure
ref: https://forum.exercism.org/t/test-failure-in-pizza-pi/1098/7

Looks like a floating point equality problem. Thought those where handled - but I guess not.

verdammelt:
The ROUNDS-TO function was used in SPACE-AGE which has the same
possible issue.

Fixes #686

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/concept/pizza-pi/pizza-pi-test.lisp
+++ b/exercises/concept/pizza-pi/pizza-pi-test.lisp
@@ -23,12 +23,16 @@
   (is (= 306 (dough-calculator 1 15)))
   (is (= 1353 (dough-calculator 5 10))))
 
+(defun rounds-to (expected actual)
+  (flet ((to-2-places (n) (/ (round (* 100 n)) 100.0)))
+    (is (= (to-2-places expected) (to-2-places actual)))))
+
 (test splash-of-sauces "Calculates the diameter of a pizza from the amount of sauce applied"
-  (is (= 32.573500793528d0 (size-from-sauce 250)))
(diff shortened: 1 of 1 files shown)


glennj:
To note the urgency of this PR: this issue blocks progress in the syllabus:

image

verdammelt:
As a note to @exercism/reviewers who may not have experience with the language

1. the issue is the use of `=` to check equality between floating point numbers - the numbers are inexact so it is problematic to check for exact equality.
2. The `rounds-to` trick is already used in`space-age` which in the past had the same sort of problem.
3. This 'trick` is simply to compare the floating point numbers only to the 2nd decimal place. as that is more than sufficient for our uses.

kotp (approved):
If the test framework has a "delta" functionality, this would explain what is happening here, that it is "close enough" to be considered equal.

verdammelt:
> If the test framework has a "delta" functionality, this would explain what is happening here, that it is "close enough" to be considered equal.

Unfortunately the testing library (`fiveam`) does not have such a utility

kotp:
> > If the test framework has a "delta" functionality, this would explain what is happening here, that it is "close enough" to be considered equal.
> 
> Unfortunately the testing library (`fiveam`) does not have such a utility

Looks like it is open source, and updates as late as 8 months ago, might be a nice addition to contribute upstream!

PR #692, opened 2022-11-29
pal-picker: A hint about how to compare strings

glennj:
## Summary

This is a good exercise, the tasks require the various conditional functions quite clearly.

However, it took me a while to crack the last task. Somehow I missed `equal` here: http://www.lispworks.com/documentation/HyperSpec/Front/X_Alph_E.htm

I see that `equality` shows up later in the syllabus. This exercise could use a hint about comparing two strings.
 
This is less of a "hint" and more like "if you're having trouble, use this function".

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/concept/pal-picker/.docs/hints.md
+++ b/exercises/concept/pal-picker/.docs/hints.md
@@ -30,3 +30,4 @@
 - This task requires single-branch conditionals
 - One action is unfitting when `pet` is a particular pet
 - The other action is unfitting unless `pet` is a particular pet
+- For comparing two strings for equality, the `equal` or `string=` functions can be used (the concept of equality is covered later in the syllabus).
--- a/exercises/concept/pal-picker/.docs/hints.md
+++ b/exercises/concept/pal-picker/.docs/hints.md
@@ -30,4 +30,7 @@
 - This task requires single-branch conditionals
 - One action is unfitting when `pet` is a particular pet
 - The other action is unfitting unless `pet` is a particular pet
-- For comparing two strings for equality, the `equal` or `string=` functions can be used (the concept of equality is covered later in the syllabus).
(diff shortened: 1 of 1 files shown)


TheLostLambda:
This is an interesting one! And probably my fault too!

When I originally wrote the exercise, I explicitly decided that things like `equal` would be [out of scope](https://github.com/exercism/common-lisp/blob/main/exercises/concept/pal-picker/.meta/design.md), but then required an understanding of it to complete the exercise!

I think that `equal` should probably remain unmentioned in this exercise, but maybe `string=` should be directly mentioned in the *instructions* for the exercise? I feel like putting something in the hints implies that it's something people should be trying to figure out when the intent of this exercise is to avoid `equality` as a concept as much as possible (to avoid stepping on the toes of the later exercise).

How would you both (@verdammelt and @glennj) feel about just adding something saying to use `string=` in the instructions for the exercise?

verdammelt:
I think adding something to the instructions sounds good. If possible maybe a forward reference to the strings concept ("...you will learn about this more in...")

glennj:
I rebased from main before pushing the latest commit. I don't understand why the rest of the recent commits from other PRs were picked up. I don't know how to clean up this PR

verdammelt:
Did you rebase off of `origin/main` or your local copy of `main` (and what that local copy up to date with `origin`)?

Locally I grabbed this branch and rebased off of `origin/main` and  now only have the top two commits.

verdammelt (approved):
Thanks for getting this PR cleaned up.

glennj:
D'oh, you were too quick! I added a commit to add a bit to the instructions:
https://github.com/glennj/exercism-common-lisp/commit/7c140a83722dcff05b02456982d979bbb61ff140

Interested in picking that up?

verdammelt:
haha - whoops!

Sure that looks ok - make a new PR and I'll approve it.

PR #693, opened 2022-11-29
high-scores: to avoid later difficulty, specify the initial score

glennj:
## Summary

For the `add-player` function, I set the player's initial score to `nil`. Reasonably I think due to the examples 
```lisp
(add-player hash-table :louis) ; => NIL
```
and the hint 
> Since the player score is not being set here, the value for the key doesn't matter

Then, the `get-score` hinted implementation
```lisp
(defun get-score (scores name) (gethash name scores 0))
```
fails one test:
```none
FAILED
Test 5
DEFAULT-SCORE (1/1)

CODE RUN
(let ((table (make-high-scores-table)))
  (add-player table :lewis)
  (is (= 0 (get-score table :lewis))))
TEST ERROR
The value
  NIL
is not of type
  NUMBER
when binding SB-KERNEL::X
```

Suggestion: when adding a player, set the initial score to zero.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/concept/high-scores/.docs/hints.md
+++ b/exercises/concept/high-scores/.docs/hints.md
@@ -7,7 +7,6 @@
 
 ## 2. Adding a player to the table
 
-- Since the player score is not being set here, the value for the key doesn't matter.
 - To add a key/value pair to a hash table use `(setf (gethash key table) value)`
 
 ## 3. Setting a player's score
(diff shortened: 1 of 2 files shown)


TheLostLambda (approved):
Another simple fix that looks great! With an already existing test-case that explicitly tests that initial scores are zero, it only makes sense to bring the docs in line!

I also think this is a bit easier and makes more sense than changing the test to match the docs. By forcing things to be initialized to zero, we're cutting down the number of different solutions / representations that we'll need to write feedback for — a win-win!

PR #695, opened 2022-11-30
fix documentation bugs for Leftovers exercise

glennj:
## Summary

Fixes broken links and removes indentation from bullet lists.

## Checklist
- [x] CI is green

### DIFF
--- a/concepts/lambda-list/about.md
+++ b/concepts/lambda-list/about.md
@@ -1,9 +1,9 @@
 # About
 
-In Common Lisp a parameter list (such as for defining a function) is also known as a (lambda list)[lambda-list]. 
+In Common Lisp a parameter list (such as for defining a function) is also known as a [lambda list][lambda-list]. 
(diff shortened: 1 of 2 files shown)


verdammelt:
@glennj Thanks for all your contributions to the track. I think for right now I am going to go with the flow and use this auto-responder that is being set up on other tracks. (I may later decide to open the track back up since it is relatively light volume...) So I guess these sorts of things should be posted in the forum.

Thanks again! I definitely appreciate your input.

glennj:
OK, no problem.

I will continue to raise issues here (that will get autoclosed) just so I can record the details for posterity. I will make a note in the forum, probably just one that will be a list of the issues found.

TheLostLambda (approved):
Hey @glennj ! Thanks for the stellar PR and cleanup!

I think we're still on that community contribution hiatus when it comes to most things, but I think small fixes like this would be nice to clear off of the track's plate :)

Thanks again!

PR #704, opened 2023-01-15
chore(exercise): update palindrome-products tests

TheLostLambda:
This brings the exercise in line with the latest version from the problem-specifications repository.

@verdammelt mentioned that our example implementation wasn't passing tests, but it seems to work for me? Perhaps I've implemented the test incorrectly? This PR definitely needs a sanity-check!

## Checklist
- [x] CI is green


verdammelt (approved):
This looks good. I could swear that when I tried this myself the existing example code did not pass the tests - but they seem to now so... [shrug].

Merge it!

PR #707, opened 2023-02-01
configlet-sync: don't run workflow on forks

fapdash:
Running this action doesn't really makes sense for forks and leads to annoying notifications for people with forks.

## Summary

This is another change I proposed for the Emacs Lisp track, maybe you like it as well.
Reference: https://github.com/exercism/emacs-lisp/pull/272

## Checklist
- [ ] CI is green

### DIFF
--- a/.github/workflows/configlet-sync.yml
+++ b/.github/workflows/configlet-sync.yml
@@ -7,6 +7,7 @@ on:
 
 jobs:
   configlet:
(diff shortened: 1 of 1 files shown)


verdammelt:
Good catch. Thanks.

verdammelt (approved):

glennj:
As a fork owner, I thank you.

PR #747, opened 2023-07-01
Fix typo and clarify a few things for `reduce` keyword arguments

colinbarry:
## Summary

Clarifies a few things with `reduce`'s keyword arguments, specifically:
- that `:from-end` takes a generalized boolean
- that `:key` is not applied to the `:initial-value`

and tidy grammar in this block.

## Checklist
- [ ] CI is green


verdammelt (approved):

verdammelt:
Thanks for fixing this!

Issue #309, opened 2021-01-28
[v3] Improve Concept Exercise Generator: add generation of concept files


verdammelt:
The common lisp concept exercise generator generates files for the exercise but not for the concepts that exercise may include.

The generator should be expanded to include a way to generate concept files.

Further it might be good if, when generating an exercise, a list of concepts can be provided which would in turn, cause those concepts to be generated if needed.

verdammelt:
This is an old issue and the old generator was removed.

However it would be still good to have a generator. It should handle creating pracitce & concept exercises - that may mean two generators?

More thought needed/wanted.

verdammelt:
I have created the beginnings of such a thing - it at least creates the stub files for concepts and concept exercises: #496 
I'll leave this issue open and as time goes on the entire issue will be finished.

verdammelt:
A practice exercise generator was added to the project - written in Python. I would still like to have a Common Lisp generator - one that can be used from the REPL.

PR #764, opened 2023-11-06
Incorporate Roswell

### LINKED ISSUE #759: Using `roswell` to run tests.
In https://github.com/exercism/common-lisp/pull/758#discussion_r1375141589, the topic of [roswell](https://github.com/roswell/roswell) was introduced.

Let's discuss it here.

glennj:
* INSTALLATION and TESTS docs updated.
* exercises/shared/.docs/tests.md updated (used to create HELP.md for downloads)
* Updated exercise generator script to create test launcher scripts.
* Added launcher scripts to all exercises

You might want to review this by commit, the "Add launcher scripts to all exercises" commit is large and repetitive.

Closes #759


glennj:
A note: the bulk of `bin/lisp_exercise_generator.py` can be replaced by a single call to `bin/configlet` -- create the director structure, create the docs/meta/tests.toml. An example, https://github.com/exercism/8th/blob/main/bin/add-exercise

verdammelt:
Sorry but I am not a fan of adding two shells scripts to every exercise. I don't think this use case is so used that it warrants the additional maintenance load.

I think editing in the web editor, for better or worse, is the main use case. The secondary use case is working in an editor and REPL.

I don't mind adding instructions on how to use roswell for others like you that wish to run the exercises from the command line, but I don't want to maintain a two shell scripts per exercise (one of which, the powershell script, I cannot even run).

Sorry if I didn't realize that this was the direction you were going; if I had realized that I would have cut it off sooner.

verdammelt:
> A note: the bulk of `bin/lisp_exercise_generator.py` can be replaced
> by a single call to `bin/configlet` -- create the director structure,
> create the docs/meta/tests.toml. An example,
> https://github.com/exercism/8th/blob/main/bin/add-exercise

Ah, this is good to know. I regret the inclusion of the python script (I had attempted to keep a "rule" of using Common Lisp for as much of the tooling as possible). It is also not a script I use. So with both those points I think it is great that we could remove it.

I'll add an issue and work on that some time later.

glennj:
> I don't mind adding instructions on how to use roswell for others like you that wish to run the exercises from the command line, but I don't want to maintain a two shell scripts per exercise

OK, no problem. I'll remove those, and revamp the docs.

> Sorry if I didn't realize that this was the direction you were going; if I had realized that I would have cut it off sooner.

That's perfectly OK. This stuff is fun.

> [removing the exercise generate script]

Do you want me to do that while I'm in here?

verdammelt:
> OK, no problem. I'll remove those, and revamp the docs.

Excellent - thanks. 

> I didn't mean to cause you undue stress with the massive commit

No problem. I understand "this stuff is fun" feeling :D .

> Do you want me to do that while I'm in here?

No. How about you add some comments to #309 (which is an old issue about writing an exercise generator).

Thanks.

verdammelt (approved):

PR #766, opened 2023-11-12
Add blank name test for two-fer

dplister:
## Summary

Within the two-fer tests, there is a test that will default the name parameter to nil, which will cause the output to be "One for , one for me." With the user intended to check for nil to ensure that "you" is substituted.

In the same pattern, a test that provides a blank input would also require the user to ensure that the output is not "One for , one for me.", with the "you" substituted.

This is a useful test for beginner developers as the usual first mitigation for a null is to default the value to empty (in this case, blank), and is a potential input to this function that the user should defend against.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/two-fer/two-fer-test.lisp
+++ b/exercises/practice/two-fer/two-fer-test.lisp
@@ -24,6 +24,8 @@
 (test another-name-given
  (is (equal "One for Bob, one for me." (two-fer:twofer "Bob"))))
 
+(test blank-name-given (is (equal "One for you, one for me." (two-fer:twofer ""))))
+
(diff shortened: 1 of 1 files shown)


verdammelt:
@dplister thank you for the pull-request. 

As you can see your new test fails given the example solution. How would you change the example solution to pass the test?

The trick bit is that two-for is the 2nd concept exercise. It already is teaching several concepts such as optional parameters, conditionals, functions and the basics of formatting. 

I don't think adding in concepts of strings and/or equality at this point would be beneficial

dplister:
@verdammelt  thanks for letting me know, happy to drop this PR.

verdammelt:
@dplister thank you for the PR submission none-the-less

PR #767, opened 2023-11-14
CI: pin GitHub Actions workflows

ErikSchierboom:
This PR updates GitHub Actions workflows to a specific version.
This ensures that the workflow will always run the same code, which makes your build _stable_.
It will also prevent a potential security issue where a tag could be replaced by a malicious commit without consumers being aware of it.

The PR updates each non-SHA based workflow reference with the SHA of the referenced version/tag, so the current behavior should not change.

See https://exercism.org/docs/building/github/gha-best-practices#h-pin-actions-to-shas for more information.

### DIFF
--- a/.github/workflows/config-checker.yml
+++ b/.github/workflows/config-checker.yml
@@ -15,7 +15,7 @@ jobs:
         uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
 
       - name: Check Config
-        uses: ./.github/actions/perform-system@{"message":"Not Found","documentation_url":"https://docs.github.com/rest"}
+        uses: ./.github/actions/perform-system
         with:
           system: config-checker
           operation: test-system
--- a/.github/workflows/config-checker.yml
+++ b/.github/workflows/config-checker.yml
@@ -12,10 +12,10 @@ jobs:
         
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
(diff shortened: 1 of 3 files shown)


.github/workflows/test-exercises.yml
```diff
@@ -12,10 +12,10 @@ jobs:
         
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
 
       - name: Run Tests
-        uses: ./.github/actions/perform-system
+        uses: ./.github/actions/perform-system@{"message":"Not Found","documentation_url":"https://docs.github.com/rest"}
```
ErikSchierboom, line 18:
```suggestion
        uses: ./.github/actions/perform-system
```

.github/workflows/config-checker.yml
```diff
@@ -12,10 +12,10 @@ jobs:
         
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11
 
       - name: Check Config
-        uses: ./.github/actions/perform-system
+        uses: ./.github/actions/perform-system@{"message":"Not Found","documentation_url":"https://docs.github.com/rest"}
```
ErikSchierboom, line 18:
```suggestion
        uses: ./.github/actions/perform-system
```

verdammelt (approved):

PR #777, opened 2024-01-07
Format track config with configlet fmt

BNAndras:
Configlet fmt can now format the track config.


verdammelt:
Hmm will have to remember to do this...

@BNAndras 
Do you think it would be useful to add this to the PR template? it already mentions `generate`...

PR #807, opened 2024-04-22
yacht: add exercise

SimaDovakin:
## Summary

For #48in24. I used the Clojure implementation as a reference.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/yacht/.meta/example.lisp
+++ b/exercises/practice/yacht/.meta/example.lisp
@@ -28,25 +28,26 @@
 
 (defun score (scores category)
   "Returns the score of the dice for the given category."
-  (cond ((equal category "ones") (count 1 scores))
-        ((equal category "twos") (* 2 (count 2 scores)))
-        ((equal category "threes") (* 3 (count 3 scores)))
-        ((equal category "fours") (* 4 (count 4 scores)))
-        ((equal category "fives") (* 5 (count 5 scores)))
-        ((equal category "sixes") (* 6 (count 6 scores)))
-        ((equal category "full house") (if (full-house (frequencies scores))
-                                           (reduce '+ scores)
-                                           0))
-        ((equal category "four of a kind") (let ((dice (four-of-a-kind-dice (frequencies scores))))
-                                                (if dice (* 4 dice) 0)))
-        ((equal category "little straight") (if (equal '(1 2 3 4 5) (sort scores #'<))
-                                                30
-                                                0))
-        ((equal category "big straight") (if (equal '(2 3 4 5 6) (sort scores #'<))
-                                             30
-                                             0))
-        ((equal category "choice") (reduce '+ scores))
-        ((equal category "yacht") (if (= 1 (hash-table-count (frequencies scores)))
-                                      50
-                                      0))
-        (t 0)))
+  (case category
+    (:ones (count 1 scores))
+    (:twos (* 2 (count 2 scores)))
+    (:threes (* 3 (count 3 scores)))
+    (:fours (* 4 (count 4 scores)))
+    (:fives (* 5 (count 5 scores)))
+    (:sixes (* 6 (count 6 scores)))
+    (:full-house (if (full-house (frequencies scores))
+                                       (reduce '+ scores)
+                                       0))
+    (:four-of-a-kind (let ((dice (four-of-a-kind-dice (frequencies scores))))
+                                            (if dice (* 4 dice) 0)))
+    (:little-straight (if (equal '(1 2 3 4 5) (sort scores #'<))
+                                            30
+                                            0))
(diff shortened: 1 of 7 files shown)


SimaDovakin:
@verdammelt Hi! Could we merge it, or should I change something?

exercises/practice/yacht/.meta/example.lisp
```diff
@@ -0,0 +1,63 @@
...
+  (let ((vals '()))
+    (progn
+      (loop for v being each hash-value of ht do (setf vals (cons v vals)))
+      vals)))
+
+(defun frequencies (sequence)
+  (let ((occurences (make-hash-table)))
+    (reduce (lambda (acc n)
+                (let ((freq (gethash n acc)))
+                  (if freq
```
verdammelt, line 16: Just a thought, and maybe I'm wrong. But what if instead of checking if `freq` is `nil` why not use a default of `0` for `gethash` so we can always do `(incf (gethash n acc 0))`.

exercises/practice/yacht/.meta/example.lisp
```diff
@@ -0,0 +1,63 @@
...
+
+(defun full-house? (scores)
+  (let* ((occurences (frequencies scores))
+         (vals (hash-values occurences)))
+        (equal '(2 3) (sort vals #'<))))
+
+(defun four-of-a-kind-dice (occurences)
+  (let ((pair nil))
+    (maphash (lambda (key val)
+              (if (>= val 4)
```
verdammelt, line 35: How about `when`?

exercises/practice/yacht/.meta/example.lisp
```diff
@@ -0,0 +1,63 @@
...
+  (cond ((equal category "ones") (count 1 scores))
+        ((equal category "twos") (* 2 (count 2 scores)))
+        ((equal category "threes") (* 3 (count 3 scores)))
+        ((equal category "fours") (* 4 (count 4 scores)))
+        ((equal category "fives") (* 5 (count 5 scores)))
+        ((equal category "sixes") (* 6 (count 6 scores)))
+        ((equal category "full house") (if (full-house? scores)
+                                           (reduce '+ scores)
+                                           0))
+        ((equal category "four of a kind") (let ((dice (four-of-a-kind-dice (frequencies scores))))
```
verdammelt, line 51: Why does `full-house?` take the scores but `four-of-a-kind-dice` take frequencies? Might be better to have consistent interface for these functions.

verdammelt (changes requested):
I think this is a good start... but I would like to discuss some small changes to make it, IMHO, better.

verdammelt:
> @verdammelt Hi! Could we merge it, or should I change something?

Sorry it took me so long to respond. But see my comments above.

SimaDovakin:
@verdammelt Thank you for review. I will improve my solution.

SimaDovakin:
@verdammelt Is it ok now?

exercises/practice/yacht/.meta/example.lisp
```diff
@@ -0,0 +1,52 @@
...
+  (let ((pair nil))
+    (maphash (lambda (key val)
+              (when (>= val 4)
+                  (setf pair key)))
+             occurences)
+    pair))
+
+(defun score (scores category)
+  "Returns the score of the dice for the given category."
+  (cond ((equal category "ones") (count 1 scores))
```
verdammelt, line 31: I think using keywords rather than strings would be more "lispy"

(I thought I had left this comment originally but it seems I had not - my apologies)

verdammelt (commented):
I think this is looking good, but using keywords rather than strings would be better.

exercises/practice/yacht/.meta/example.lisp
SimaDovakin: Am I right @verdammelt? I should use instead of categories as strings (e.g., "ones", "twos"), categories as keywords (e.g., :ones, :twos).

verdammelt: Yes, I believe that for this sort of usage, where the values are effectively constants, flags, "magic values" that keywords would be more idiomatic. (`"four of a kind"` can become `:four-of-a-kind`)

So instead of `"ones"` use `:ones`. And since keywords are unique you can then use `eq` to compare them. `case` might also work well here rather than `cond`.

SimaDovakin: I got it. Thank you! I'll change it.

SimaDovakin:
@verdammelt Now categories are keywords, and there is a `case` instead of `cond` in the main function.

verdammelt (approved):
Thank you for taking the time for all the updates.

SimaDovakin:
Thank you for all your reviews!

PR #847, opened 2025-01-14
Fix test for Common Lisp - Anagram

brasse:
Make the test ignore order of the anagram sets returned.

## Summary

(Explain what this change is attempting to accomplish)

## Checklist
- [ ] CI is green

### DIFF
--- a/exercises/practice/anagram/anagram-test.lisp
+++ b/exercises/practice/anagram/anagram-test.lisp
@@ -15,12 +15,15 @@
 ;; Define and enter a new FiveAM test-suite
 (def-suite* anagram-suite)
 
+(defun set-equal-p (set0 set1)
+  (null (set-exclusive-or set0 set1 :test #'equal)))
+
 (test no-matches
   (is (equal nil
              (anagram:anagrams-for "diaper" '("hello" "world" "zombies" "pants")))))
 
 (test detects-two-anagrams
-  (is (equal '("lemons" "melons")
+  (is (set-equal-p '("lemons" "melons")
              (anagram:anagrams-for "solemn" '("lemons" "cherry" "melons")))))
 
 (test does-not-detect-anagram-subsets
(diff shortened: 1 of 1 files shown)


tasxatzial:
Forum thread

https://forum.exercism.org/t/whats-the-process-to-follow-if-i-find-an-exercise-where-the-instructions-are-wrong-or-misleading/14816

verdammelt (changes requested):
Solution looks good.

Looks like the instructions of the problem changed about 3 years ago and the tests never updated.

Of course comparing two lists which have no reason to *necessarily* be in the same order is of course a dumb mistake (one I make in my tests over and over again [sigh]). I usually simply compare two sorted lists, but here since the language has the nicely built in `set-exclusive-or` it seems like a good idea.

verdammelt (approved):
Previous review got marked 'request changes' by accident. Sorry for the confusion.

verdammelt:
> Forum thread
> 
> https://forum.exercism.org/t/whats-the-process-to-follow-if-i-find-an-exercise-where-the-instructions-are-wrong-or-misleading/14816

Before merging I'll check back on this thread over the next couple days to see who wins the "the track is wrong" vs. "the problem statement is wrong" argument.

IMHO, regardless of if the problem statement said 'list' or 'set' the tests are in the wrong comparing two lists which we have no reason to believe would _necessarily_ be sorted in the same order.

exercises/practice/anagram/anagram-test.lisp
```diff
@@ -15,12 +15,15 @@
 ;; Define and enter a new FiveAM test-suite
 (def-suite* anagram-suite)
 
+(defun set-equal-p (set0 set1)
+  (null (set-exclusive-or set0 set1 :test #'equal)))
```
verdammelt, line 19: After some thought perhaps `#'string-equal` would be good here? As type-checking/documentation that these are unordered collections of strings?

verdammelt (approved):

exercises/practice/anagram/anagram-test.lisp
verdammelt: Having reconsidered a little - I think #'equal is a reasonable choice.

PR #848, opened 2025-01-15
exercises/practice/anagram: add an append with instructions about the order of the return values

IsaacG:


verdammelt:
@IsaacG is this new description actually representative of what this track currently does?

If Exercism(TM) has decided that the exercise is to check that the right _set_ of anagrams is found (i.e. unordered collection) then in my opinion this track should use the PR #847 which ensures the tests do not accidentally test for any specific ordering.

Am I correct that the decision (from this thread: https://forum.exercism.org/t/anagram-exercise-instructions-and-tests-dont-align-on-ordering/14816/48) is that the problem by itself (without appended text) be for an unordered collection?

BNAndras:
My understanding is that once we come to a consensus, the new instructions won't imply whether an ordered or unordered collection is expected so the student isn't influenced one way or the other. To that end, the append clarifies what the track expects.  So that other PR should be resolved first, and then the append in this PR can be updated to reflect what the track now expects. However, you can also choose not to merge this PR.

IsaacG:
Clarify what this track expects. Context: https://forum.exercism.org/t/14816/

IsaacG:
What @BNAndras said. The current instructions are ambiguous. Does "a set of anagrams" in prose mean a `set` object containing the anagrams? Or is "set" meant in the colloquial meaning, ie a collection/group? Many tracks do not have a `set` object. Most tracks expect a `list` object, not a `set` object.

The problem spec is being updated to avoid using the word "set" to remove that confusion ... and learning the implementation details unspecified. Tracks can use any data type they prefer. Some tracks expect a comma-separated string. Most expect a list. Some expect a set. Some expect text to be written to a filehandle (STDOUT). C expects a connection of objects to be updated, where each object has a Boolean `is_anagram` value.

The append file allows tracks to (optionally) explicitly state expectations. The append file is certainly not required. It is supposed to reflect the existing track tests (if I read them right).

verdammelt:
I plan on merging #847 so I don't think we want/need this change anymore

PR #854, opened 2025-03-20
Correct mistake in optional parameters concept documentation

### LINKED ISSUE #842: Lilly's Lasagna Leftovers: two small issues
**Issue 1**. This is a learning exercise, so it has an **Introduction** devoted to lambda lists. In the **Optional Parameters** section, there is a line
```
(optional-parameters 5 nil 10) ; => (5 DEFAULT (:Z-WAS-SUPPLIED 10))
```
In fact, it should be:
```
(optional-parameters 5 nil 10) ; => (5 NIL (:Z-WAS-SUPPLIED 10))
```
That is, `nil` value for a parameter is not the same as not supplying a value for the parameter.

Additionally, it's worth adding to the **Introduction** a note about how to correctly skip a default parameter (well, it must be a keyword parameter for us to be able to do that).

**Issue 2**. This exercise continues another learning exercise: _Lilly's Lasagna_. _Lilly's Lasagna Leftovers_ is meant to re-implement some functions from _Lilly's Lasagna_ in a more sophisticated way. The design in _Lilly's Lasagna_ was: `expected-time-in-oven` defines total cooking time; `remaining-minutes-in-oven` subtracts from it `elapsed-time-in-oven` to produce its result. In _Lilly's Lasagna Leftovers_, the total cooking time is calculated in a more sophisticated way, but the function that is required to be rewritten is `remaining-minutes-in-oven` (which doesn't have an `elapsed-time-in-oven` parameter now).

Imo, the preferable way of fixing this issue is to rewrite `expected-time-in-oven` rather than `remaining-minutes-in-oven`. Regarding the latter, one small change can be: the parameter `elapsed-time-in-oven` should now have a default value of 0.

verdammelt:
Fixes #842

## Checklist
- [X] If docs where changed run `./bin/configlet generate` to ensure all documents are properly generated.
- [x] CI is green

[no important files changed]

### DIFF
--- a/concepts/optional-parameters/introduction.md
+++ b/concepts/optional-parameters/introduction.md
@@ -13,5 +13,5 @@ Finally a symbol can be specified for each optional parameter which will be boun
                              (list :z-was-not-supplied z))))
 
 (optional-parameters)          ; => (NIL DEFAULT (:Z-WAS-NOT-SUPPLIED NIL))
-(optional-parameters 5 nil 10) ; => (5 DEFAULT (:Z-WAS-SUPPLIED 10))
+(optional-parameters 5 nil 10) ; => (5 NIL (:Z-WAS-SUPPLIED 10))
 ```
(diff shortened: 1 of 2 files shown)


BNAndras (approved):

verdammelt:
Thanks @BNAndras for the review

PR #855, opened 2025-03-20
Replace `beer-song` with `bottle-song`

### LINKED ISSUE #768: Deprecate `beer-song` in favor of `bottle-song`
See https://github.com/exercism/problem-specifications/blob/main/exercises/beer-song/.deprecated. bottle-song will need to be ported before we deprecate beer-song. I can work on that by the end of this month if there are no takers before then.

verdammelt:
This PR adds `bottle-song` and deprecates `beer-song`.

Fixes #768

### DIFF
--- a/exercises/practice/bottle-song/.meta/config.json
+++ b/exercises/practice/bottle-song/.meta/config.json
@@ -1,0 +1,19 @@
+{
+  "authors": [
+    "verdammelt"
+  ],
+  "files": {
+    "solution": [
+      "./bottle-song.lisp"
+    ],
+    "test": [
+      "./bottle-song-test.lisp"
+    ],
+    "example": [
+      "./.meta/example.lisp"
+    ]
+  },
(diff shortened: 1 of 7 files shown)


exercises/practice/bottle-song/.meta/config.json
```diff
@@ -0,0 +1,17 @@
+{
+  "authors": [],
```
BNAndras, line 2: Missing your GH username here

verdammelt:
@BNAndras I went ahead implemented a version of `bottle-song` which would fix your issue #768.

Do you think this is a reasonable solution?

BNAndras (approved):

exercises/practice/bottle-song/.meta/config.json
verdammelt: Thanks! fixed now.

PR #857, opened 2025-03-26
Fix bottle-song config file

verdammelt:
Config file generated with `configlet` sets paths to start with `"./"`
which appears to cause problems in the UI.


verdammelt:
Closing as duplicate of #856

PR #860, opened 2025-04-07
Add `reverse-string` exercise

wsgac:
## Summary

I noticed the absence of this exercise from the CL track. I found it instructive and kinda tricky when dealing with the combining characters in strings.

## Checklist
- [x] CI is green

### DIFF
--- a/exercises/practice/reverse-string/.docs/instructions.md
+++ b/exercises/practice/reverse-string/.docs/instructions.md
@@ -1,0 +1,5 @@
+# Instructions
+
+Reversing strings (reading them from right to left, rather than from left to right) is a surprisingly common task in programming.
+
+For example, in bioinformatics, reversing the sequence of DNA or RNA strings is often important for various analyses, such as finding complementary strands or identifying palindromic sequences that have biological significance.
--- a/exercises/practice/reverse-string/.meta/config.json
+++ b/exercises/practice/reverse-string/.meta/config.json
@@ -1,0 +1,15 @@
+{
+   "title": "Reverse String",
+   "blurb": "Reverse a given string.",
+   "source": "Introductory challenge to reverse an input string",
+   "source_url": "https://medium.freecodecamp.org/how-to-reverse-a-string-in-javascript-in-3-different-ways-75e4763c68cb",
+   "deep_dive_youtube_id": "biGAfK6OElE",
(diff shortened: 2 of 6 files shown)


wsgac:
@verdammelt I was trying to write in the forum using the link above, but unfortunately there's an error when I try to log in. Would you consider reopening my PR?

verdammelt:
My first question was: why not just use `reverse`?

I see your solution uses a SBCL specific function - what happens when the student is not using SBCL?

If we can formulate the solution in such a way to not use non-standard functions then we can move forward toward inclusion.

In a quick check it looks like `reverse` works correctly except perhaps for the final test where instead of the exact expected string my REPL printed out something that looks like it should be equivalent as the final character is preceded by two unicode characters which add the top and bottom accent marks to the final character.

wsgac:
Well, `reverse` works on individual characters (even Unicode ones), but it fails when dealing with combining characters. You can see this when `coerce`ing one of the test strings - "Würstchenstand". In this case the `ü` is not the character `#\LATIN_SMALL_LETTER_U_WITH_DIAERESIS`, but rather a composition of `#\u` and `#\COMBINING_DIAERESIS`. So in order to properly reverse, we need to maintain the order of combining characters. In other words, we need to reverse groups composed of proper characters followed by zero or more modifier characters that combine together.

As for the portability, the only portable way I know of that actually worked for me was using `cl-unicode:bidi-class`, but that would mean depending on an external library. This problem is particularly pronounced with the Thai language test (`grapheme-clusters`). In that example we have a single letter with two modifiers with names not suggesting that they combine at all: `(#\THAI_CHARACTER_PHO_PHUNG #\THAI_CHARACTER_SARA_UU #\THAI_CHARACTER_MAI_THO)`. Any thoughts on that?

verdammelt:
If there is no way to do this without pulling in external dependencies or using implementation specific functionality I don't think we should include it.

While my opinion on this may not matter much - but this track has always stuck to *only* the functionality provided by the language according to the standard.

BNAndras:
You could add the exercise but not include specific tests that are problematic. Some students might choose to use the built-in reverse, but surely others will try to implement the reversing functionality on their own.

wsgac:
Or those two tests could be guarded by a feature reader macro (e.g. `#+include-combining-characters`) and users instructed to push that into `*features*` if they want to try.

PR #861, opened 2025-04-14
Add `resistor-color`

BNAndras:
The huge diff in the track config is because I used `configlet create --practice-exercise` which formats that file. The only relevant bit in that mess is that I gave resistor-color a difficulty of 2.

I can add `resistor-color-duo` next to continue the resistor color series of exercises.

### DIFF
--- a/exercises/practice/resistor-color/.meta/example.lisp
+++ b/exercises/practice/resistor-color/.meta/example.lisp
@@ -5,20 +5,8 @@
 
 (in-package :resistor-color)
 
-(defparameter *color-codes*
-  '(("black" . 0)
-    ("brown" . 1)
-    ("red" . 2)
-    ("orange" . 3)
-    ("yellow" . 4)
-    ("green" . 5)
-    ("blue" . 6)
-    ("violet" . 7)
-    ("grey" . 8)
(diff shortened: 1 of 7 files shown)


verdammelt:
I created a separate PR for the config file formatting - if we merge that then you can rebase off it and have a cleaner PR - sound good?

Otherwise I like this PR and will approve once cleaned up.

exercises/practice/resistor-color/.meta/example.lisp
```diff
@@ -0,0 +1,24 @@
+(defpackage :resistor-color
+  (:use :cl)
+  (:export :color-code
+           :colors))
+
+(in-package :resistor-color)
+
+(defparameter *color-codes*
```
verdammelt, line 8: Could also just be a list of the colors and then use `position`?

BNAndras: Yeah, that's reasonable, and I'll update the example. Normally I don't refactor my example solutions since they're not public-facing (for practice exercises at least), but I'll make a note to do it for future exercises as well.

verdammelt (approved):

PR #864, opened 2025-04-16
Add `square-root`

BNAndras:
I had trouble running the exercise generator for this one (see below), so I had to manually generate the files. The example is adapted from my TypeScript implementation for this exercise.

```
(base) ➜  exercism-common-lisp git:(main) python ./bin/lisp_exercise_generator.py 
Enter the path (relative or absolute) to the problem-specifications repository: /home/anagy/Documents/GitHub/exercism-problem-specifications 
Enter the name of the exercise you wish to generate: square-root
You are about to overwrite an existing exercise!  Confirm (Y/N): y
Author's Github handle: BNAndras
Traceback (most recent call last):
  File "/home/anagy/Documents/GitHub/exercism-common-lisp/./bin/lisp_exercise_generator.py", line 465, in <module>
    no_arguments()
  File "/home/anagy/Documents/GitHub/exercism-common-lisp/./bin/lisp_exercise_generator.py", line 410, in no_arguments
    brand_new_exercise(exercise_name, prob_spec_exercise, author)
  File "/home/anagy/Documents/GitHub/exercism-common-lisp/./bin/lisp_exercise_generator.py", line 330, in brand_new_exercise
    create_instructions(exercise_name, prob_spec_exercise)
  File "/home/anagy/Documents/GitHub/exercism-common-lisp/./bin/lisp_exercise_generator.py", line 73, in create_instructions
    with open(input_file, 'r', encoding = "utf-8") as read_from:
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
FileNotFoundError: [Errno 2] No such file or directory: '/home/anagy/Documents/GitHub/exercism-problem-specifications/exercises/square-root/description.md'
```

### DIFF
--- a/config.json
+++ b/config.json
@@ -1315,6 +1315,14 @@
         "difficulty": 1
       },
       {
+        "slug": "square-root",
+        "name": "Square Root",
+        "uuid": "bb0ee09d-5d5c-45d6-949b-b65b04a5f5e7",
+        "practices": [],
+        "prerequisites": [],
+        "difficulty": 2
+      },
+      {
         "slug": "resistor-color",
         "name": "Resistor Color",
         "uuid": "59c87423-c430-4e46-9790-bef60c9c62eb",
--- a/exercises/practice/square-root/.docs/instructions.md
+++ b/exercises/practice/square-root/.docs/instructions.md
@@ -1,0 +1,18 @@
+# Instructions
+
+Your task is to calculate the square root of a given number.
+
+- Try to avoid using the pre-existing math libraries of your language.
+- As input you'll be given a positive whole number, i.e. 1, 2, 3, 4…
+- You are only required to handle cases where the result is a positive whole number.
+
+Some potential approaches:
+
+- Linear or binary search for a number that gives the input number when squared.
+- Successive approximation using Newton's or Heron's method.
+- Calculating one digit at a time or one bit at a time.
+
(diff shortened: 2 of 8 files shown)


verdammelt:
I don't know much about that exercise generator, maybe I should get rid of it.

As for this exercise - the example code seems to go into an infinite loop for a non-perfect-square.. say 5. It appears that all the test cases are only perfect squares; but the instructions imply that the input could be any positive integer.

Shall we add more test cases? Or at least make the example file do something reasonable?

BNAndras:
The instructions also specify the returned value will always be a positive integer so the student wouldn’t need to handle the square root of 5. We can certainly do an append noting that isn’t true on this track and then add additional tests to that effect. Then the example would need to be updated to pass the additional tests.

BNAndras:
43 tracks implement the exercise as-is with the canonical test cases. Julia, R, and V each added a track-specific test. Julia and R added an input validation test for negative radicands, and V added a test for the square root for 5764801 which is 2401. None of those three added an append, possibly because the added tests don't change the existing instructions.

For the TypeScript solution, I only loop ten times, and that's enough for those tests. For an input like 5, it'd give the wrong answer, but it wouldn't hang. I originally had that same ten loops in this example, but I removed it since it wasn't needed for the current tests. I can add it back though.

verdammelt:
If you could easily add some termination clause (and have the method to evaluate to `nil`) that would be great. But If not, then let's merge it as is. I will approve.

verdammelt (approved):

PR #865, opened 2025-04-17
Add `resistor-color-trio`

BNAndras:
This is the last of the canonical resistor-color exercises. If you're interested in adding it, Python and Crystal added their own [Resistor Color Expert](https://github.com/exercism/python/blob/main/exercises/practice/resistor-color-expert/.docs/instructions.md) which adds a tolerance value as well.


verdammelt:
`(error (format nil "Value ~a is too large to be formatted" value)`

can be simplified to 

`(error "Value ~a is too large to be formatted" value)`

`ERROR` takes care of calling `FORMAT` on its arguments.

BNAndras:
Thanks, I didn’t realize that.

verdammelt:
Also it looks like the example.lisp could use a quick reformatting - some indentation is a bit off.

verdammelt (approved):

PR #866, opened 2025-04-18
Add `house`

BNAndras:


verdammelt (changes requested):
This looks good but the stub file has some non-lisp-y symbol names: `startVerse` and `endVerse`. 

I am going to approve - but please fix those before actually merging. 

Thanks.

verdammelt (approved):

BNAndras:
That’s from the problem specs. Did you have something in mind? Something basic like start and end?

verdammelt:
> That’s from the problem specs. Did you have something in mind? Something basic like start and end?

Sure - or maybe `start-verse` and `end-verse` - just name them in a more lisp-y fashion.

PR #867, opened 2025-04-25
Update files array in config.json

BNAndras:
## Summary

Unless the paths are manually fixed during exercise generation, the exercise can’t be solved on the website.

https://forum.exercism.org/t/the-new-exercise-of-common-lisp-has-bugs/16505/11

### DIFF
--- a/config.json
+++ b/config.json
@@ -20,16 +20,16 @@
   },
   "files": {
     "solution": [
-      "./%{kebab_slug}.lisp"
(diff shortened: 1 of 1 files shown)


verdammelt (approved):
I could swear that I did this when a previous exercise did not work (the bottles-song). but I guess I only fixed it in the generated exercise rather than here. Thank you.

PR #868, opened 2025-04-25
Fix house file paths

BNAndras:
## Summary

The exercise is currently not solvable due to the file paths being incorrect when loaded into the online editor. A separate PR fixes that in the track config.

https://forum.exercism.org/t/the-new-exercise-of-common-lisp-has-bugs/16505/19


verdammelt (approved):
What about the other recent exercises you've done - are their paths ok? (the resistor-color exercises...)

PR #872, opened 2025-05-04
Add atbash-cipher decode tests

### LINKED ISSUE #871: Atbash Cipher should have tests for decoding.
The current tests of "Atbash Cipher" in Common Lisp has no tests for the decode function. As the canonical test data has a few of them, I think that they should be implemented.

verdammelt:
These tests were never implemented (but were listed in the tests.toml
file as if they had been).

Fixes #871


verdammelt:
@BNAndras I found it rather simple (and a nice distraction) to implement these missing tests. Could you give them a quick review? Thanks.

BNAndras (approved):

PR #877, opened 2025-05-05
Sync `leap` tests

BNAndras:
## Summary

I found some more tests to sync. There are 752 completions to date for this exercise, and the ones that passed the old test suite would probably still pass the new one. You can probably get away with not rerunning those solutions on the updated tests. Up to you.


verdammelt:
@BNAndras I agree - let's not have the tests rerun. Add `[no important files changed]` to your merge message.

verdammelt (approved):

PR #878, opened 2025-05-05
Sync `scrabble-score` tests

BNAndras:
## Summary

This syncs the exercise's canonical tests from upstream. There are only 130 completions, but I'm not sure that it's worth rerunning them. The new tests don't add anything new.


verdammelt (commented):
Agree that the additional tests do not seem like they add much. Let's not have all the existing submissions retested.

BNAndras:
I don’t think you approved the PR with that review.

verdammelt (approved):

verdammelt:
> I don’t think you approved the PR with that review.

weird, sorry about that.

PR #880, opened 2025-05-05
Update `rna-transcription` test names

BNAndras:
## Summary

All canonical tests are implemented, but I noticed minor typos in the test names. There's no need to rerun tests here.


verdammelt (approved):
thanks for correcting test names, will make maintenance easier.

PR #881, opened 2025-05-06
Sync `meetup` tests

BNAndras:
## Summary

It seems a good chunk of these tests were present, but at least a handful for different years were missing. I also did a good bit of cleanup swapping around parameter order to match the previous example.


verdammelt (approved):
Thanks for this update - this is a VERY OLD exercise!

PR #887, opened 2025-05-12
add `relative-distance`

BNAndras:


exercises/practice/relative-distance/.meta/example.lisp
```diff
@@ -0,0 +1,54 @@
...
+            (let ((parent (car entry))
+                  (children (cdr entry)))
+
+              (unless (gethash parent neighbors)
+                (setf (gethash parent neighbors) (make-hash-table :test 'equal)))
+
+              (dolist (child children)
+                      (unless (gethash child neighbors)
+                        (setf (gethash child neighbors) (make-hash-table :test 'equal)))
+                      (setf (gethash child (gethash parent neighbors)) t)
```
verdammelt, line 20: `setf` can take pairs of symbols and values so instead of `(setf a b) (setf c d)` you can do `(setf a b c d)`. I personally think it looks nicer instead of two expressions.

exercises/practice/relative-distance/.meta/example.lisp
```diff
@@ -0,0 +1,54 @@
...
+    (unless (and (gethash person-a neighbors) (gethash person-b neighbors))
+      (return-from degree-of-separation nil))
+
+
+    (let ((queue (list (cons person-a 0)))
+          (visited (make-hash-table :test 'equal)))
+
+      (setf (gethash person-a visited) t)
+
+      (loop while queue do
```
verdammelt, line 38: Starting to review the function I was thinking: "oh `dolist` not `loop` nice..." but now that I see `loop` I wonder why you didn't use `loop` above as well. You could do something like: `(loop for (parent . child) in tree ...)`

I don't care `do` vs. `loop` but think it would good to be consistent.

exercises/practice/relative-distance/.meta/example.lisp
```diff
@@ -0,0 +1,54 @@
...
+                   (current-person (car current))
+                   (current-degree (cdr current)))
+
+              (when (equal current-person person-b)
+                (return-from degree-of-separation current-degree))
+
+              (let ((current-neighbors (gethash current-person neighbors)))
+                (when current-neighbors
+                  (maphash (lambda (neighbor _)
+                             (unless (gethash neighbor visited)
```
verdammelt, line 49: Don't you get a `STYLE-WARNING` here about the unused symbol `_`? Use `(declare (ignore _))` Maybe give it a reasonable name even though it is ignored...

exercises/practice/relative-distance/.meta/example.lisp
```diff
@@ -0,0 +1,54 @@
...
+                               (setf (gethash neighbor visited) t)
+                               (setf queue (append queue (list (cons neighbor (1+ current-degree)))))))
+                           current-neighbors))))))
+
+    nil))
```
verdammelt, line 54: LOL -with the length of this function I've forgotten what condition this `NIL` might be from! Especially now in the GitHub Diff view with comments splitting up the code :D

verdammelt (changes requested):
I think some cleanup would be good. I have left comments about them.

exercises/practice/relative-distance/.meta/example.lisp
verdammelt: Ah, I now see where this Is for. I'll leave a separate comment about it.

exercises/practice/relative-distance/.meta/example.lisp
```diff
@@ -0,0 +1,54 @@
...
+
+      (setf (gethash person-a visited) t)
+
+      (loop while queue do
+            (let* ((current (pop queue))
+                   (current-person (car current))
+                   (current-degree (cdr current)))
+
+              (when (equal current-person person-b)
+                (return-from degree-of-separation current-degree))
```
verdammelt, line 44: How about you factor this `loop` so it evaluates to `current-degree` and then this whole function can evaluate to what this `loop` evaluates to?

verdammelt (changes requested):

verdammelt:
Just a general comment - the use of several `return-from` expressions feels a little odd to me - I called out one in `loop` but the one in the middle of the function also could be replace by using `when` rather than `unless` right?

BNAndras:
Sorry, I’ve been a bit distracted, but I’ll pick this back up in a couple of days.

verdammelt:
Thank you for coming back and finishing this up. I gave it  a quick look and it seems fine so I am approving. I don't want to hold it up anymore with any nit-picking I may be prone to do.

PR #911, opened 2026-06-03
Fix function name from 'feeding-time-p' to 'feeding-time'

Maartz:
## Summary

As it's not a predicate, do not return T or NIL we should ditch the `-p` suffix.

## Checklist
- [ ] CI is green


verdammelt:
I see that you opened 3 PRs to do this one change. I do like this change (good catch) but it should be a single PR. Do all the changes in this PR (I will close the others)

verdammelt (changes requested):
Must complete this change - all affected files, not just documentation.

verdammelt:
Seems like the original author is ignoring this PR so I am closing it. I will to the work myself as I think it is a reasonable change.

PR #913, opened 2026-06-03
Update pal-picker.lisp

Maartz:
## Summary

As it's not a predicate, do not return T or NIL we should ditch the `-p` suffix.

## Checklist
- [ ] CI is green


verdammelt:
Should not be a separate PR - make this change in #911

PR #915, opened 2026-07-08
Remove unused pathname and structure references from equality concept

### LINKED ISSUE #467: Improve equality concept
If possible work the concept & exercise such that this concept can be placed earlier in the track. Right now it depends upon strings and characters and arrays etc.

Perhaps with forward references to other concepts these pre-requisites can be removed.

RohanKapri:
## Summary

This PR removes outdated references from the equality concept documentation.

### Changes
- Remove pathname description and example from `about.md`
- Remove structure description and example from `about.md`
- Remove structure description and example from `introduction.md`

Fixes #467


verdammelt:
@RohanKapri I'm not sure what you mean by "outdated". As far as I know these documents have the correct information. Is it because these concepts are not present on the track?

RohanKapri:
Hi @verdammelt,

Thanks for the review.

Yes, that was my intention. I interpreted issue #467 as suggesting that the references to structures and pathnames could be removed because those concepts are not introduced or used in the current track at this point.

If I misunderstood the intent of the issue, I'm happy to update the PR or make whatever changes you think are appropriate.

Thanks!

verdammelt:
I had forgotten about that old issue. I've had the thought before and maybe I should go through with it: remove all those old issues! I'm definitely NOT going to be working on them any time soon and they should all be re-thought.

As for this change: I don't think I would like to include it. I don't think the documentation is wrong to include info about features not currently covered by the track.

Thank you for taking the time and interest to make the change. But I will not merge them in at this time.

